PR #3, opened 2016-11-09
Use correct artifact name for plugin

mr-archano:
The artifact name specified in the `publish` configuration is not following the accepted internal convention. Maven coordinates for the artifact should be
```
com.novoda:gradle-static-analysis-plugin:0.1
```


blundell:
👍  no tests to update this time?

mr-archano:
No, tests are not pulling the plugin from the maven repo. Also note that the plugin is not deployed anywhere yet.

PR #4, opened 2016-11-09
Update the website in the publish configuration

mr-archano:
The plugin has moved out of `novoda/spikes`, but the publish configuration was still pointing to that repo.

> Hopefully this is the last change before the first release to Bintray 😅


blundell:
:)

PR #6, opened 2016-11-10
Don't override default effort and report level

eduardb:
> Tracked in JIRA by [PT-312](https://novoda.atlassian.net/browse/PT-312)

### Scope of the PR

The effort and report level of FindBugs were enforced by us, overriding the defaults, but this is something that a user might not want. Leave the defaults in place, and let the user choose if he wants other values.

### Considerations/Implementation Details

Really straight forward, just remove setting the values in the task configurator. 

Since by default FindBugs is not that restrictive anymore,  `LowPriorityViolator.java` was updated to contain a violation with priority 3 that will get picked up with the defaults.

#### Testing

Updated (fix) test that said that it should fail build when threshold is passed

Added a test to showcase how one can change the effort and report level

### DIFF
--- a/plugin/src/test/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsIntegrationTest.groovy
+++ b/plugin/src/test/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsIntegrationTest.groovy
@@ -43,6 +43,22 @@ class FindbugsIntegrationTest {
     }
 
     @Test
+    public void shouldDetectMoreWarningsWhenEffortIsMaxAndReportLevelIsLow() {
+        TestProject.Result result = projectRule.newProject()
+                .withSourceSet('debug', SOURCES_WITH_LOW_VIOLATION, SOURCES_WITH_MEDIUM_VIOLATION)
+                .withPenalty('''{
+                    maxErrors = 0
+                    maxWarnings = 1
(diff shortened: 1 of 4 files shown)


plugin/src/test/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsIntegrationTest.groovy
```diff
@@ -31,18 +31,34 @@ class FindbugsIntegrationTest {
...
 
     @Test
+    public void shouldDetectMoreWarningsWhenEffortIsMaxAndReportLevelIsLow() {
+        TestProject.Result result = projectRule.newProject()
+                .withSourceSet('debug', SOURCES_WITH_LOW_VIOLATION, SOURCES_WITH_MEDIUM_VIOLATION)
+                .withPenalty('''{
+                    maxErrors = 0
+                    maxWarnings = 1
+                }''')
+                .withFindbugs('findbugs { effort = \'max\' \n reportLevel = \'low\'}')
```
mr-archano: you could use `"` instead of escaping `'`

eduardb: Done in 0bd326a

PR #8, opened 2016-11-11
Improve exclude rules

mr-archano:
> Tracked in JIRA by [PT-313](https://novoda.atlassian.net/browse/PT-313)

### Scope of the PR

While integrating v0.1 of the plugin in one of our projects we realised that the `exclude` directive in the plugin extension is only supporting [file patterns](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/util/PatternFilterable.html), but there is no way to specify an entire source set for instance.

### Considerations/Implementation Details

Introduced `SourceFilter` as thin delegate holding all the exclude rules to be applied to a specific `SourceTask`. File patterns are supported as well as `File`s or collections of them, eg:
```gradle
exclude project.fileTree('src/test/java')
exclude project.file('src/main/java/foo/bar/Constants.java')
exclude project.sourceSets.main.java.srcDirs
exclude '**/*Test.java'
```

#### Testing
`SourceFilter` is fully tested to cover all the scenarios above. A bunch of integration tests have been added for each configurator to validate the `exclude` functionality in both Java and Android projects.

> Thanks to His Groovyness @devisnik for the help with the refactoring of the `SourceFilter`. It looks waaaaay better than expected ✨ .

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/SourceFilter.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/SourceFilter.groovy
@@ -1,24 +1,43 @@
 package com.novoda.staticanalysis.internal
 
+import org.gradle.api.Project
 import org.gradle.api.file.FileCollection
+import org.gradle.api.file.FileTree
 import org.gradle.api.tasks.SourceTask
 
 class SourceFilter {
 
+    private final Project project
     private final List<Object> excludes = []
 
-    void exclude(Object rule) {
-        excludes.add(rule)
+    SourceFilter(Project project) {
+        this.project = project
+    }
+
+    void exclude(Object exclude) {
+        excludes.add(exclude)
     }
 
     void applyTo(SourceTask task) {
-        excludes.each { filter ->
-            if (filter instanceof FileCollection) {
-                task.source = task.source.findAll { !filter.contains(it) }
+        excludes.each { exclude ->
+            if (exclude instanceof File) {
+                apply(task, project.files(exclude))
+            } else if (exclude instanceof FileCollection) {
+                apply(task, exclude)
+            } else if (exclude instanceof Iterable<File>) {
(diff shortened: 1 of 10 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/internal/SourceFilter.groovy
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    void applyTo(SourceTask task) {
+        excludes.each { exclude ->
+            if (exclude instanceof File) {
+                apply(task, project.files(exclude))
+            } else if (exclude instanceof FileCollection) {
+                apply(task, exclude)
+            } else if (exclude instanceof Iterable<File>) {
+                apply(task, exclude.inject(null, accumulateIntoTree()) as FileTree)
```
frapontillo: So is `inject` some kind of `fold`?

plugin/src/main/groovy/com/novoda/staticanalysis/internal/SourceFilter.groovy
```diff
@@ -0,0 +1,43 @@
...
+            }
+        }
+    }
+
+    private void apply(SourceTask task, FileCollection excludedFiles) {
+        task.source = task.source.findAll { !excludedFiles.contains(it) }
+    }
+
+    private def accumulateIntoTree() {
+        return { tree, file -> tree?.plus(project.fileTree(file)) ?: project.fileTree(file) }
```
frapontillo: HAX so much

plugin/src/main/groovy/com/novoda/staticanalysis/internal/SourceFilter.groovy
mr-archano: yep. why `inject` instead of `fold` or the more popular `reduce` is beyond me :/

PR #16, opened 2017-01-27
PT-420: Honour Android variants in Chesktyle and PMD

mr-archano:
> Tracked by [PT-240](https://novoda.atlassian.net/browse/PT-420)

## Scope of the PR

During the integration of this plugin in one of our running projects @xrigau noticed that all Checkstyle and PMD tasks were run even tho some of the variants were disabled. The root issue is in the way we are creating such tasks, where only the source sets are evaluated without actually considering the variants. 

- Introduced generics to correctly enforce type for the `TestProject` builders
- Used `ApplicationVariant`/`TestVariant`/`UnitTestVariant` to determine which source sets should be inspected by Checkstyle and PMD.
- If the same source set is used in multiple variants (eg: `main`) we ensure only one `CheckstyleMain`/`PmdMain` task is created

### Test(s) added 

Added integration tests checking behaviour of Checkstyle and PMD wrt Android variants. Scenarios include:
- violations in test variants (android and unit test)
- no product flavors
- multiple product flavors
- manually disabled variants (via `variantFileter`)

### Paired with 
@ 👻

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/pmd/PmdConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/pmd/PmdConfigurator.groovy
@@ -45,18 +45,21 @@ class PmdConfigurator extends CodeQualityConfigurator<Pmd, PmdExtension> {
     @Override
     protected void configureAndroid(Object variants) {
         project.with {
-            android.sourceSets.all { sourceSet ->
-                def sourceDirs = sourceSet.java.srcDirs
-                def notEmptyDirs = sourceDirs.findAll { it.list()?.length > 0 }
-                if (notEmptyDirs.empty) {
-                    return
-                }
-                Pmd pmd = tasks.create("pmd${sourceSet.name.capitalize()}", Pmd)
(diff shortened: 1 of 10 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/internal/pmd/PmdConfigurator.groovy
```diff
@@ -45,18 +45,21 @@ class PmdConfigurator extends CodeQualityConfigurator<Pmd, PmdExtension> {
...
-                pmd.with {
-                    description = "Run PMD analysis for ${sourceSet.name} classes"
-                    source = sourceSet.java.srcDirs
-                }
-                variants.all { variant ->
+            variants.all { variant ->
+                variant.sourceSets.each { sourceSet ->
+                    def taskName = "pmd${sourceSet.name.capitalize()}"
+                    Pmd pmd = tasks.findByName(taskName)
+                    if (pmd == null) {
```
tasomaniac: I like the `null` check. If I understood it right, this ensures that it is created only once.

PR #20, opened 2017-02-03
PT-427: Make Findbugs to honour the exclude filters

mr-archano:
> Tracked by [PT-427](https://novoda.atlassian.net/browse/PT-427)

## Scope of the PR

Findbugs is currently ignoring the exclude filters specified as part of the plugin configuration.
The issue is related to the fact the tool is working against class files while the exclude filters are applied only to sources. To solve the problem we need to filter out the classes produced from source files excluded in the config.

We decided to create include patterns for the `classes` property of the `Findbugs` task evaluating the filtered set of source files specified via `source` (and after the `exclude` filters from the plugin extension have been applied), for instance:

- original source file: `com.novoda.test.HighPriorityViolator.java`
- generated include pattern: `com/novoda/test/HighPriorityViolator*`

Note that the `*` at the end will allow to catch inner classes as well.
Incidentally the internals of `*Configurator` classes have been refactored to expose a clean configuration point for Java projects via `configureJavaProject()`.

### Test(s) added 

`FindbugsIntegrationTest` is now testing the case where we provide an `exclude` filter to skip one specific file from the sources as opposed to the entire source set as before. The source file used in the test is also now using a proper package and contains an inner class, to make the test more significative.

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
@@ -4,9 +4,14 @@ import com.novoda.staticanalysis.EvaluateViolationsTask
 import com.novoda.staticanalysis.internal.CodeQualityConfigurator
 import org.gradle.api.Action
 import org.gradle.api.Project
+import org.gradle.api.file.ConfigurableFileTree
+import org.gradle.api.file.FileCollection
 import org.gradle.api.plugins.quality.FindBugs
 import org.gradle.api.plugins.quality.FindBugsExtension
+import org.gradle.api.tasks.SourceSet
 import org.gradle.internal.logging.ConsoleRenderer
+
+import java.nio.file.Path
 
 class FindbugsConfigurator extends CodeQualityConfigurator<FindBugs, FindBugsExtension> {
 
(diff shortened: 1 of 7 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
```diff
@@ -45,23 +50,71 @@ class FindbugsConfigurator extends CodeQualityConfigurator<FindBugs, FindBugsExt
...
+                    .findAll { Path sourceDir -> sourceFile.startsWith(sourceDir) }
+                    .collect { Path sourceDir -> sourceDir.relativize(sourceFile) }
+        }
+        .flatten()
+    }
+
+    /**
+     * The simple "classes = sourceSet.output" may lead to non-existing resources directory
+     * being passed to FindBugs Ant task, resulting in an error
+     * */
```
devisnik: Is above comment still applicable?

plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
```diff
@@ -45,23 +50,71 @@ class FindbugsConfigurator extends CodeQualityConfigurator<FindBugs, FindBugsExt
...
             }
         }
     }
 
+    private static List<String> createIncludesPatternsFrom(FileCollection sourceFiles, List<File> sourceDirs) {
+        List<Path> includedSourceFiles = sourceFiles.matching { '**/*.java' }.files.collect { it.toPath() }
+        createIncludePatterns(includedSourceFiles, sourceDirs.collect { it.toPath() })
+    }
+
+    private static List<String> createIncludePatterns(List<Path> includedSourceFiles, List<Path> sourceDirs) {
```
devisnik: NIT: Here it's `...Include...`, in the method name above `...Includes...`.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/pmd/PmdConfigurator.groovy
```diff
@@ -43,32 +43,28 @@ class PmdConfigurator extends CodeQualityConfigurator<Pmd, PmdExtension> {
...
+                    if (task == null) {
+                        task = tasks.create(taskName, Pmd)
+                        task.with {
+                            description = "Run PMD analysis for ${sourceSet.name} classes"
+                            source = sourceSet.java.srcDirs
                         }
                     }
-                    pmd.mustRunAfter variant.javaCompile
+                    sourceFilter.applyTo(task)
+                    task.mustRunAfter variant.javaCompile
```
devisnik: Is the `mustRunAfter` still needed here?

plugin/src/test/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsIntegrationTest.groovy
```diff
@@ -134,13 +134,13 @@ class FindbugsIntegrationTest {
...
-                .withSourceSet('debug', SOURCES_WITH_LOW_VIOLATION, SOURCES_WITH_MEDIUM_VIOLATION)
-                .withSourceSet('release', SOURCES_WITH_HIGH_VIOLATION)
+                .withSourceSet('debug', SOURCES_WITH_LOW_VIOLATION)
+                .withSourceSet('release', SOURCES_WITH_HIGH_VIOLATION, SOURCES_WITH_MEDIUM_VIOLATION)
                 .withPenalty('''{
                     maxErrors = 0
                     maxWarnings = 10
                 }''')
-                .withFindbugs('findbugs { exclude "HighPriorityViolator.java" }')
+                .withFindbugs('findbugs { exclude "com/novoda/test/HighPriorityViolator.java" }')
```
devisnik: 👍

plugin/src/main/groovy/com/novoda/staticanalysis/internal/pmd/PmdConfigurator.groovy
devisnik: Might be to ensure the compiler fails first, even before the codestyle checks run, right?

mr-archano: I'm actually not sure this `mustRunAfter` is correct here to be honest - like we briefly discussed last week, I reckon we should consider the use of `dependsOn` instead.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
mr-archano: Well, we are not changing the default behaviour here, instead we stack filters upon the default implementation: this method has been copied (and groovyfied) from the original [`FindbugsPlugin#L189-L191`](https://github.com/gradle/gradle/blob/master/subprojects/code-quality/src/main/groovy/org/gradle/api/plugins/quality/FindBugsPlugin.java#L189-L191). Should we reference the source?

PR #23, opened 2017-02-15
PT-434: Better classes filtering for Findbugs tasks

mr-archano:
> Tracked by [PT-434](https://novoda.atlassian.net/browse/PT-434)

## Scope of the PR

While integrating the plugin in one of our projects we realised that Findbugs was analysing a source folder even though an exclude rule was provided for it.

After quite some investigation we realised that the issue is related to the way* we create include patterns from the filtered collection in `source`. When the source is empty (eg: all the source files have been filtered out) no include pattern can be generated, therefore the collection of classes won't be filtered at all. This sometimes is not a problem because the analysis task is skipped anyway (`source` property is marked as [`@SkipWhenEmpty`](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/SourceTask.html#getSource())), but this doesn't seem enough to cover all cases.
We decided then to enforce an empty collection for `classes` when an empty collection is found in `source`.

Incidentally the task generating html reports for findbugs is now skipped when the related analysis has not run.

### Test(s) added 

`FindbugsIntegrationTest` now contains few more tests to check:
- which `:findbug*` task has been executed/skipped
- whether an empty `classes` property is provided to each `findbugs*` task with an empty `source`

------

> (*) Given a `Findbugs` task we evaluate the filtered set of source files and for each of them we create an include rule for the related class(es).
>Let's assume source contains `foo/Bar.java`; we expect the plugin will produce `foo/Bar*` as include pattern to allow the `Bar.class` (and its inner classes) to be included in
the analysis.

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
@@ -62,10 +62,14 @@ class FindbugsConfigurator extends CodeQualityConfigurator<FindBugs, FindBugsExt
             sourceFilter.applyTo(task)
             task.conventionMapping.map("classes", {
                 List<String> includes = createIncludePatterns(task.source, androidSourceDirs)
-                project.fileTree(variant.javaCompile.destinationDir).include(includes)
-            }.memoize());
+                getAndroidClasses(variant, includes)
+            });
             task.dependsOn variant.javaCompile
         }
+    }
+
+    private FileCollection getAndroidClasses(Object variant, List<String> includes) {
+        includes.isEmpty() ? project.files() : project.fileTree(variant.javaCompile.destinationDir).include(includes)
     }
 
     @Override
@@ -78,8 +82,8 @@ class FindbugsConfigurator extends CodeQualityConfigurator<FindBugs, FindBugsExt
                 task.conventionMapping.map("classes", {
                     List<File> sourceDirs = sourceSet.allJava.srcDirs.findAll { it.exists() }.toList()
                     List<String> includes = createIncludePatterns(task.source, sourceDirs)
-                    createClassesTreeFrom(sourceSet).include(includes)
-                }.memoize());
+                    getJavaClasses(sourceSet, includes)
+                });
             }
         }
     }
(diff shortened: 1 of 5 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
```diff
@@ -62,12 +62,16 @@ class FindbugsConfigurator extends CodeQualityConfigurator<FindBugs, FindBugsExt
             sourceFilter.applyTo(task)
             task.conventionMapping.map("classes", {
                 List<String> includes = createIncludePatterns(task.source, androidSourceDirs)
-                project.fileTree(variant.javaCompile.destinationDir).include(includes)
-            }.memoize());
```
mr-archano: During the investigation of the bug I realised that the `classes` property is evaluated multiple times during the configuration phase of an Android project, so I decided to avoid possible side-effects of memoization.

plugin/src/test/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsIntegrationTest.groovy
```diff
@@ -202,4 +230,129 @@ class FindbugsIntegrationTest {
...
+                    exclude project.android.sourceSets.androidTest.java.srcDirs
+                }''')
+                .withAdditionalConfiguration(addCheckFindbugsClassesTask())
+                .build('checkFindbugsClasses')
+
+        Truth.assertThat(result.outcome(':checkFindbugsClasses')).isEqualTo(TaskOutcome.SUCCESS)
+    }
+
+    /**
+     * <p>While integrating the plugin in one of our projects we realised that Findbugs was analysing a source folder
```
ouchadam: as great as this information is, do you think the first paragraph helps?

It's an awesome PR description though!

mr-archano: No, I don't think it adds more insight really. I wanted to put some explanation on why that (gross) task exists tho, maybe I can shave all the context and leave only the hint about the check?

plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/GenerateHtmlReport.groovy
```diff
@@ -13,11 +13,9 @@ class GenerateHtmlReport extends JavaExec {
...
-        if (xmlReportFile != null && xmlReportFile.exists()) {
-            main = 'edu.umd.cs.findbugs.PrintingBugReporter'
-            standardOutput = new FileOutputStream(htmlReportFile)
-            args '-html', xmlReportFile
-            super.exec()
-        }
+        main = 'edu.umd.cs.findbugs.PrintingBugReporter'
+        standardOutput = new FileOutputStream(htmlReportFile)
+        args '-html', xmlReportFile
+        super.exec()
```
devisnik: 👍

PR #25, opened 2017-02-16
PT-442: Customise reports base url in logs

mr-archano:
> Tracked by [PT-442](https://novoda.atlassian.net/browse/PT-442)

## Scope of the PR

When running static analyses the output always refers to local file URLs for reports which is fine on a local machine but unusable for CI logs.

- Introduced `ReportUrlRenderer` as delegate to render a report URL in the logs. Default implementation will render URLs using `ConsoleRenderer` from Gradle.
- Introduced `LogsExtension` to allow users to specify custom renderers as simple closure `(File) -> String`. `LogsExtension` exposes a `reportBaseUrl` as convenience method that will allow to swap the default base url of reports with one of choice.

### Test(s) added 

`LogsConfigurationIntegrationTest` is now checking basic configuration of `LogsExtension` including usage built-in convenience methods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ It's possible to specify a custom renderer for the report urls in the logs via t
 staticAnalysis {
     ...
     logs {
-        baseReportUrl "http://ci.mycompany.com/job/myproject/ws/app/build/reports"
+        reportBaseUrl "http://ci.mycompany.com/job/myproject/ws/app/build/reports"
     }
 }
 ```
--- a/README.md
+++ b/README.md
@@ -76,6 +76,25 @@ the relevant html reports, for instance:
     - file:///foo/project/build/reports/pmd/main4.html
 ```
 
+It's possible to specify a custom renderer for the report urls in the logs via the `logs` extension. This can be useful in CI
+ environments, where the local paths are not reachable directly. For instance the snippet below will replace the base url with
+ one of your choice:
+```gradle
+staticAnalysis {
+    ...
+    logs {
+        baseReportUrl "http://ci.mycompany.com/job/myproject/ws/app/build/reports"
+    }
+}
+```
+so that in the logs you will see the report urls printed as
+```
+> Checkstyle rule violations were found (0 errors, 1 warnings). See the reports at:
+- http://ci.mycompany.com/job/myproject/ws/app/build/reports/checkstyle/main.html
+```
+More info on the topic can be found in the [`LogsExtension`](https://github.com/novoda/gradle-static-analysis-plugin/blob/master/plugin/src/main/groovy/com/novoda/staticanalysis/LogsExtension.groovy)
+groovydoc.
+
 #### Out-of-the-box support for Android projects
 Android projects use a gradle model that is not compatible with the Java one, supported by the built-in static analysis tools plugins.
 Applying `gradle-static-analysis-plugin` to your Android project will make sure all the necessary tasks are created and correctly configured
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/EvaluateViolationsTask.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/EvaluateViolationsTask.groovy
@@ -1,15 +1,15 @@
 package com.novoda.staticanalysis
 
 import com.novoda.staticanalysis.internal.Violations
-import org.gradle.api.DefaultTask
 import org.gradle.api.GradleException
 import org.gradle.api.NamedDomainObjectContainer
+import org.gradle.api.internal.ConventionTask
 import org.gradle.api.tasks.TaskAction
-import org.gradle.internal.logging.ConsoleRenderer
 
-class EvaluateViolationsTask extends DefaultTask {
+class EvaluateViolationsTask extends ConventionTask {
     private PenaltyExtension penaltyExtension
     private NamedDomainObjectContainer<Violations> violationsContainer
+    private ReportUrlRenderer reportUrlRenderer
 
     EvaluateViolationsTask() {
         group = 'verification'
@@ -26,6 +26,10 @@ class EvaluateViolationsTask extends DefaultTask {
 
     Violations maybeCreate(String name) {
         violationsContainer.maybeCreate(name)
+    }
+
+    ReportUrlRenderer getReportUrlRenderer() {
+        reportUrlRenderer
     }
 
     @TaskAction
@@ -50,10 +54,9 @@ class EvaluateViolationsTask extends DefaultTask {
         }
     }
 
-    static String getViolationsMessage(Violations violations) {
-        def consoleRenderer = new ConsoleRenderer()
+    String getViolationsMessage(Violations violations) {
         "$violations.name rule violations were found ($violations.errors errors, $violations.warnings warnings). See the reports at:\n" +
-                "${violations.reports.collect { "- ${consoleRenderer.asClickableFileUrl(it)}" }.join('\n')}"
+                "${violations.reports.collect { "- ${reportUrlRenderer.render(it)}" }.join('\n')}"
     }
 
 }
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/EvaluateViolationsTask.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/EvaluateViolationsTask.groovy
@@ -5,6 +5,7 @@ import org.gradle.api.DefaultTask
 import org.gradle.api.GradleException
 import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.tasks.TaskAction
+import org.gradle.internal.logging.ConsoleRenderer
 
 class EvaluateViolationsTask extends DefaultTask {
     private PenaltyExtension penaltyExtension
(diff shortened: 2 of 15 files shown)


README.md
```diff
@@ -76,6 +76,25 @@ the relevant html reports, for instance:
...
 ```
 
+It's possible to specify a custom renderer for the report urls in the logs via the `logs` extension. This can be useful in CI
+ environments, where the local paths are not reachable directly. For instance the snippet below will replace the base url with
+ one of your choice:
+```gradle
+staticAnalysis {
+    ...
+    logs {
+        reportBaseUrl "http://ci.mycompany.com/job/myproject/ws/app/build/reports"
```
stefanhoth: Not sure this is the right example path for a base url. Shouldn't end at `.../ws/`?

plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisPlugin.groovy
```diff
@@ -20,10 +20,11 @@ class StaticAnalysisPlugin implements Plugin<Project> {
     }
 
     private EvaluateViolationsTask createEvaluateViolationsTask(Project project) {
-        StaticAnalysisExtension extension = project.extensions.create('staticAnalysis', StaticAnalysisExtension)
+        StaticAnalysisExtension extension = project.extensions.create('staticAnalysis', StaticAnalysisExtension, project)
```
stefanhoth: this looks odd. Is there no common way to access the project within the new extension?

mr-archano: An extension is just an object, so you have to pass a reference of a project somehow. I don't see how this is odd :(

stefanhoth: No offense! I found it odd that you're referencing a project `project.extensions.create(...` and then need to pass in the project itself again for inner workings. Dunno, seems repetitive. But what do I know. 😉

README.md
mr-archano: The example follows the implementation of `reportBaseUrl` as convenience method built on top of the more generic `reportUrlRenderer` facility, so it's correct. The emphasis is put on the reports location rather than the project location as you are suggesting.
I'm happy to remove this utility and/or change the sample, but having `reportBaseUrl` to change only the project base url sounds wrong to me. We can easily add another `projectBaseUrl` convenience method too.

stefanhoth: I was just worried they don't match up because in my mind we previously only talked about the base url as the workspace. If the reports folder is the same for all tools and that's the base I'm super happy with this example.

plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisPlugin.groovy
mr-archano: Well, what I'm trying to say is this is Gradle API: extensions are created by reflection via [`project.extensions.create(String, Class, java.lang.Object...)`](https://docs.gradle.org/current/javadoc/org/gradle/api/plugins/ExtensionContainer.html#create(java.lang.String, java.lang.Class, java.lang.Object...)), and they don't get further properties injected.

README.md
mr-archano: Yes, all reports are collected into `build/reports` as per Gradle convention. The same folder is also used to put test reports or code coverage reports too.

plugin/src/main/groovy/com/novoda/staticanalysis/EvaluateViolationsTask.groovy
```diff
@@ -1,14 +1,15 @@
...
 
 import com.novoda.staticanalysis.internal.Violations
-import org.gradle.api.DefaultTask
 import org.gradle.api.GradleException
 import org.gradle.api.NamedDomainObjectContainer
+import org.gradle.api.internal.ConventionTask
 import org.gradle.api.tasks.TaskAction
 
-class EvaluateViolationsTask extends DefaultTask {
+class EvaluateViolationsTask extends ConventionTask {
```
tasomaniac: Is this really necessary? We used conventionMapping with extending `DefaultTask`
I wanted to comment because it is in `internal` package.

plugin/src/test/groovy/com/novoda/staticanalysis/internal/LogsConfigurationIntegrationTest.groovy
```diff
@@ -0,0 +1,88 @@
...
+import static com.novoda.test.LogsSubject.assertThat
+
+@RunWith(Parameterized.class)
+class LogsConfigurationIntegrationTest {
+
+    private static
+    final String DEFAULT_CHECKSTYLE_CONFIG = "checkstyle { configFile new File('${Fixtures.Checkstyle.MODULES.path}') }"
+
+    @Parameterized.Parameters
+    public static List<Object[]> rules() {
```
tasomaniac: Can't you just have `List` here? Does it have to be List of Object array?
Not 100% sure but I would expect following to work.

```
@Parameterized.Parameters
public static List rules() {
    [TestProjectRule.forJavaProject(), TestProjectRule.forAndroidProject()]
}
```

plugin/src/main/groovy/com/novoda/staticanalysis/LogsExtension.groovy
```diff
@@ -0,0 +1,58 @@
...
+    private ReportUrlRenderer reportUrlRenderer = ReportUrlRenderer.DEFAULT
+    final Project project
+
+    LogsExtension(Project project) {
+        this.project = project
+    }
+
+    /**
+     * This method will create a new {@code ReportUrlRenderer} using the transformation closure provided.
+     * the closure will receive the report as {@code File} and should provide a {@code String} as output.
```
devisnik: Capital `T` ?

plugin/src/main/groovy/com/novoda/staticanalysis/LogsExtension.groovy
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * Here an example os custom renderer that will render report urls as markdown links (?!):
```
devisnik: type `os`

plugin/src/main/groovy/com/novoda/staticanalysis/LogsExtension.groovy
```diff
@@ -0,0 +1,58 @@
...
+     * <pre>
+     *     staticAnalysis {
+     *         ...
+     *         logs {
+     *             reportUrlRenderer { report -> "$[report.name]($report.path)" }
```
devisnik: Shouldn't this be `[$` ?

plugin/src/main/groovy/com/novoda/staticanalysis/LogsExtension.groovy
```diff
@@ -0,0 +1,58 @@
...
+            String render(File report) {
+                renderer.call(report)
+            }
+        }
+    }
+
+    /**
+     * This method provides a convenient way for creating a custom {@code ReportUrlRenderer} that will replace the base
+     * url of each report in the logs with one of choice. The target for the replacement by default is set to the
+     * reports dir of the project, but can be changes if needed.
```
devisnik: typo `changes` -> `changed` ?

plugin/src/main/groovy/com/novoda/staticanalysis/LogsExtension.groovy
```diff
@@ -0,0 +1,58 @@
...
+     *
+     * @param newBaseUrl the base url to be used
+     * @param localBaseUrl optional, the local base url to be replaced with {@code newBaseUrl}
+     */
+    void reportBaseUrl(String newBaseUrl, String localBaseUrl = "$project.buildDir/reports") {
+        reportUrlRenderer { report -> newBaseUrl + (report.path - ~/$localBaseUrl/) }
```
devisnik: Wouldn't `newBaseUrl + (report.path - localBaseUrl)` do the same here?

plugin/src/main/groovy/com/novoda/staticanalysis/EvaluateViolationsTask.groovy
mr-archano: I understand your concern about the `internal` package, but I used `ConventionTask` because it's extending `DefaultTask`, not the other way around. Maybe Gradle works its magic and adds a convention map to any task anyway? I can definitely try to keep `DefaultTask` here

plugin/src/test/groovy/com/novoda/staticanalysis/internal/LogsConfigurationIntegrationTest.groovy
mr-archano: I'll try, I was of the idea that simple `List` were not supported by all version of Junit. Let me check if we're good.

plugin/src/main/groovy/com/novoda/staticanalysis/LogsExtension.groovy
mr-archano: It actually depends on `localBaseUrl` I guess: we want to ensure the regex is applied to the `report.path` instead of the whole `newBaseUrl + report.path`.

PR #27, opened 2017-02-20
PT-273/PT-450: Add support for rules as Maven artifact

mr-archano:
> Tracked by [PT-450](https://novoda.atlassian.net/browse/PT-450)

## Scope of the PR
We would like to provide a way to consume the same set of rules cross-projects for all the static analysis tools we intend to use. A viable way to share such rules is as maven artifact (JAR).

Added a container of `RulesExtension` to `StaticAnalysisExtension` that can be used to define artifacts containing rules we want to use to configure the static analysis tools, eg:
```
staticAnalysis {
  rules {
    novoda {
      maven 'foo.bar:static-analysis-rules:0.1'
    }
  }
  checkstyle {
    config rules.novoda['checkstyle/modules.xml']
  }
}
```
where `foo.bar:static-analysis-rules:0.1` is a JAR deployed to a maven repository of choice. The  configuration XML file can be accessed via the `getAt` operator providing the relative path inside the JAR. It is then retrieved as `TextResource` that can be fed - for instance - to `checkstyle.config` or anywhere such type is supported.

### Test(s) added 
Added `RulesIntegrationTest` to check the new feature. To help with the setup we introduced a `DeployRulesTestRule` in order to create and then deploy a set of rules as JAR artifact into a local maven repo that can be used in the integration test.

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisExtension.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisExtension.groovy
@@ -1,6 +1,8 @@
 package com.novoda.staticanalysis
 
 import org.gradle.api.Action
+import org.gradle.api.NamedDomainObjectContainer
+import org.gradle.api.NamedDomainObjectFactory
 import org.gradle.api.Project
 
 class StaticAnalysisExtension {
@@ -22,9 +24,16 @@ class StaticAnalysisExtension {
 
     private PenaltyExtension currentPenalty = new PenaltyExtension()
     private final LogsExtension logs
+    private final NamedDomainObjectContainer<RulesExtension> rules
 
     StaticAnalysisExtension(Project project) {
         this.logs = new LogsExtension(project)
+        this.rules = project.container(RulesExtension, new NamedDomainObjectFactory<RulesExtension>() {
+            @Override
+            RulesExtension create(String name) {
+                new RulesExtension(name, project)
+            }
+        })
     }
 
     void penalty(Action<? super PenaltyExtension> action) {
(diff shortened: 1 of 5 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisExtension.groovy
```diff
@@ -22,9 +24,16 @@ class StaticAnalysisExtension {
 
     private PenaltyExtension currentPenalty = new PenaltyExtension()
     private final LogsExtension logs
+    private final NamedDomainObjectContainer<RulesExtension> rules
 
     StaticAnalysisExtension(Project project) {
         this.logs = new LogsExtension(project)
+        this.rules = project.container(RulesExtension, new NamedDomainObjectFactory<RulesExtension>() {
```
takecare: fancy 👍

plugin/src/test/groovy/com/novoda/test/DeployRulesTestRule.groovy
```diff
@@ -0,0 +1,92 @@
...
+
+sourceSets {
+    main {
+        resources {
+            srcDirs = ${resourceDirs.collect { "'$it.path'" }}
+        }
+    }
+}
+
+apply plugin: 'maven-publish'
```
mr-archano: Note: I went for the [_New Maven Publishing_](https://docs.gradle.org/current/userguide/publishing_maven.html) support. The old one I was used to has a problem with the generation of the metadata/pom file on the first run.

plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisExtension.groovy
tasomaniac: So we are expecting multiple rules to be defined. Makes sense.

mr-archano:
@tasomaniac I thought about that, but it would be weird to read in the `README` of an unreleased feature. I plan to amend the documentation as part of the preparation for next release.

PR #28, opened 2017-02-21
PT-447: Explicitly include Android variants in analysis

mr-archano:
> Tracked by [PT-447](https://novoda.atlassian.net/browse/PT-447)

## Scope of the PR
When applying the plugin we want to skip the creation of tasks for certain Android variants to speed up the whole analysis. After discussing this with @ouchadam and @devisnik we agreed it would be cool to have an explicit way to tell the plugin which Android variants should be considered and which shouldn't.

Each `*Configurator` now decorates the tool specific extension with an `includeVariants` method that will receive a predicate used to decide which variant should be considered as part of the analysis. Eg:
```
staticAnalsysis {
  findbugs {
    includeVariant { variant -> variant.name.equals('debug') }
  }
}
```
By default - for the moment - all variants will be included unless an explicit predicate is provided via `includeVariants`.

### Test(s) added 
Few integration tests have been added to validate that tasks related to excluded variants are not created by the plugin. Incidentally some redundant integration test has been removed because their scenarios were already covered.

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
@@ -14,12 +14,14 @@ abstract class CodeQualityConfigurator<T extends SourceTask, E extends CodeQuali
     protected final Violations violations
     protected final EvaluateViolationsTask evaluateViolations
     protected final SourceFilter sourceFilter
+    protected Closure<Boolean> includeVariantsFilter
 
     protected CodeQualityConfigurator(Project project, Violations violations, EvaluateViolationsTask evaluateViolations) {
         this.project = project
         this.violations = violations
         this.evaluateViolations = evaluateViolations
         this.sourceFilter = new SourceFilter(project)
+        this.includeVariantsFilter = { true }
     }
 
     void execute() {
@@ -28,6 +30,7 @@ abstract class CodeQualityConfigurator<T extends SourceTask, E extends CodeQuali
             project.extensions.findByType(extensionClass).with {
                 defaultConfiguration.execute(it)
                 ext.exclude = { Object rule -> sourceFilter.exclude(rule) }
+                ext.includeVariants = { Closure<Boolean> filter -> includeVariantsFilter = filter }
                 config.delegate = it
                 config()
             }
@@ -35,9 +38,11 @@ abstract class CodeQualityConfigurator<T extends SourceTask, E extends CodeQuali
                 boolean isAndroidApp = project.plugins.hasPlugin('com.android.application')
                 boolean isAndroidLib = project.plugins.hasPlugin('com.android.library')
                 if (isAndroidApp || isAndroidLib) {
-                    configureAndroidProject(isAndroidApp ? project.android.applicationVariants : project.android.libraryVariants)
-                    configureAndroidProject(project.android.testVariants)
-                    configureAndroidProject(project.android.unitTestVariants)
+                    NamedDomainObjectSet<Object> variants = project.container(Object)
+                    variants.addAll(isAndroidApp ? project.android.applicationVariants : project.android.libraryVariants)
+                    variants.addAll(project.android.testVariants)
+                    variants.addAll(project.android.unitTestVariants)
+                    configureAndroidProject(variants.matching { includeVariantsFilter(it) })
                 } else {
                     configureJavaProject()
                 }
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
@@ -3,6 +3,7 @@ package com.novoda.staticanalysis.internal
 import com.novoda.staticanalysis.EvaluateViolationsTask
 import com.novoda.staticanalysis.StaticAnalysisExtension
 import org.gradle.api.Action
+import org.gradle.api.NamedDomainObjectSet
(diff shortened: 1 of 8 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
```diff
@@ -27,16 +30,19 @@ abstract class CodeQualityConfigurator<T extends SourceTask, E extends CodeQuali
...
                 config()
             }
             project.afterEvaluate {
                 boolean isAndroidApp = project.plugins.hasPlugin('com.android.application')
                 boolean isAndroidLib = project.plugins.hasPlugin('com.android.library')
                 if (isAndroidApp || isAndroidLib) {
-                    configureAndroidProject(isAndroidApp ? project.android.applicationVariants : project.android.libraryVariants)
-                    configureAndroidProject(project.android.testVariants)
-                    configureAndroidProject(project.android.unitTestVariants)
+                    NamedDomainObjectSet<Object> variants = project.container(Object)
```
tasomaniac: Why not using just `Set`?

mr-archano: I actually tried, but it was failing to execute `addAll()` with a cryptic stacktrace that sounded like a covariance issue.
The rationale of using `NamedDomainObjectSet` is that every `project.android.*Variants` is already of this type, and it supports an elegant `matching()` that reads better than the `findAll()` provided in `Collection`.

tasomaniac: Makes sense. `matching` is definitely looks better.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
```diff
@@ -64,7 +70,7 @@ abstract class CodeQualityConfigurator<T extends SourceTask, E extends CodeQuali
         }
     }
 
-    protected abstract void configureAndroidProject(Object variants)
+    protected abstract void configureAndroidProject(NamedDomainObjectSet variants)
```
ouchadam: 👍

plugin/src/test/groovy/com/novoda/staticanalysis/internal/checkstyle/CheckstyleAndroidVariantIntegrationTest.groovy
```diff
@@ -11,30 +12,15 @@ import static com.novoda.test.LogsSubject.assertThat
 
 class CheckstyleAndroidVariantIntegrationTest {
 
-    private static final String DEFAULT_CHECKSTYLE_CONFIG = "checkstyle { configFile new File('${Fixtures.Checkstyle.MODULES.path}') }"
+    private static final String DEFAULT_CHECKSTYLE_CONFIG = """
+        checkstyle { configFile new File('${Fixtures.Checkstyle.MODULES.path}') }
+    """
```
ouchadam: how come these are `"""`?

plugin/src/test/groovy/com/novoda/staticanalysis/internal/checkstyle/CheckstyleAndroidVariantIntegrationTest.groovy
```diff
@@ -110,60 +103,105 @@ class CheckstyleAndroidVariantIntegrationTest {
...
+                """)
                 .build('check')
 
-        assertThat(result.logs).doesNotContainCheckstyleViolations()
+        assertThat(result.logs).doesNotContainLimitExceeded()
+        assertThat(result.logs).containsCheckstyleViolations(1, 1,
+                result.buildFileUrl('reports/checkstyle/main.html'),
+                result.buildFileUrl('reports/checkstyle/demo.html'))
+        def checkstyleTasks = result.tasksPaths.findAll { it.startsWith(':checkstyle') }
+        Truth.assertThat(checkstyleTasks).containsAllIn([
```
ouchadam: awhhh yeahh 💯

plugin/src/test/groovy/com/novoda/staticanalysis/internal/checkstyle/CheckstyleAndroidVariantIntegrationTest.groovy
mr-archano: because of the string interpolation (ie: `${...}`), you need to use `GString` (quoted in `""`) instead of simple `String` (using `''`)

plugin/src/test/groovy/com/novoda/staticanalysis/internal/checkstyle/CheckstyleAndroidVariantIntegrationTest.groovy
mr-archano: If this doesn't make you and @eduardb happy I really don't know what will :D

plugin/src/test/groovy/com/novoda/staticanalysis/internal/checkstyle/CheckstyleAndroidVariantIntegrationTest.groovy
ouchadam: ahh didn't notice the inner `''` 👍

mr-archano:
The reason why you see these strings is because how the integration tests have been designed: each integration tests creates a proper gradle project, and the strings provided in the configuration are collated in their `build.gradle`. Such test-projects are then run via the Gradle tooling API so we can check the outputs (reports, logs, tasks, build success, etc).

Issue #11, opened 2016-12-13
Should rerunning the task report previous failures?


ouchadam:
Given a previously failed result, when running the task again the build does not fail. 

Should we consider this expected?

mr-archano:
@ouchadam I'm not sure I understanding what you're asking. To clarify: the build will fail **iff** the thresholds are violated. Running the same task again on the same sources should lead to the same result. Do you have a sample where we can reproduce this issue?

ouchadam:
The static analysis task had cached it's result which meant it was skipped the second time the build was ran. This meant the build passed even though there were violations.

mr-archano:
OH. That's nasty. I would say this behaviour is probably inherited from one of the static analysis tools integrated (findbugs?) but we can try to workaround it once we create our tasks.
to be investigated (I'll add it to our backlog).

mr-archano:
I tried this with different setups but couldn't reproduce. Closing the issue for now. @ouchadam if you encounter this issue again please let me know.

mr-archano:
I talked too early: it seems I found a project where I can reproduce this. I will share more info shortly. Reopening for now.

mr-archano:
I could reproduce the issue easily using an integration test. The problem seems related to the fact the violations are now collected only if static analysis tasks (eg: `checkstyleMain`) are run. When one of these tasks is `UP-TO-DATE` because inputs are the same then we don't evaluate the reports (already generated) and this yields a wrong result for `evaluateViolations`.
:man_facepalming:
One way of fixing this is to move the collection of violations as separate task that would run always. This could be combined with the feature requested in #18 and lead to better configuration API altogether.

mr-archano:
Fixed in #30

PR #30, opened 2017-03-19
Bugfix: Ensure multiple runs invariant

mr-archano:
## Scope of the PR
As @ouchadam mentioned in issue #11:
> Given a previously failed result, when running the task again the build does not fail.

This shouldn't happen because code is identical, therefore should yield the same results. As already mentioned in #11, after a quick investigation:
> The problem seems related to the fact the violations are now collected only if static analysis tasks (eg: `checkstyleMain`) are run. When one of these tasks is `UP-TO-DATE` because inputs are the same then we don't evaluate the reports (already generated) and this yields a wrong result for `evaluateViolations`.

Each analysis task (`Checkstyle`, `Findbugs`, `Pmd`) is now depending on a specific implementation of `CollectionViolationsTask` that will perform the same action before provided as `doLast`. This will allow to evaluate the reports and collect the violations even when the analysis are skipped because their input haven't changed.

### Test(s) added 
For each tool supported by the plugin an integration test has been added to ensure that 2 consecutive runs without any change in the code yield the same result.


plugin/src/main/groovy/com/novoda/staticanalysis/internal/CollectViolationsTask.groovy
```diff
@@ -0,0 +1,37 @@
...
+
+    CollectViolationsTask() {
+        onlyIf { xmlReportFile?.exists() }
+    }
+
+    void setXmlReportFile(File xmlReportFile) {
+        this.xmlReportFile = xmlReportFile
+    }
+
+    void setViolations(Violations violations) {
```
blundell: heh this reminds me of the PMD source code

plugin/src/main/groovy/com/novoda/staticanalysis/internal/CollectViolationsTask.groovy
```diff
@@ -0,0 +1,37 @@
...
+    final void run() {
+        collectViolations(xmlReportFile, htmlReportFile, violations)
+    }
+
+    File getXmlReportFile() {
+        return xmlReportFile
+    }
+
+    File getHtmlReportFile() {
+        new File(xmlReportFile.absolutePath - '.xml' + '.html')
```
blundell: cool syntax

plugin/src/test/groovy/com/novoda/staticanalysis/internal/checkstyle/CheckstyleIntegrationTest.groovy
```diff
@@ -44,6 +44,29 @@ public class CheckstyleIntegrationTest {
     }
 
     @Test
+    public void shouldFailBuildAgainWhenCheckstyleWarningsStillOverTheThresholdAfterSecondRun() {
```
blundell: 👍 

might be nice to word it differently tho

`afterSecondRunWhenStillOverTheThresholdThenTheBuildFailsAgain`

or if you insist on starting with should:

`shouldFailTheBuildAfterSecondRunWhenStillOverTheThreshold`

plugin/src/main/groovy/com/novoda/staticanalysis/internal/pmd/CollectPmdViolationsTask.groovy
```diff
@@ -0,0 +1,22 @@
...
+
+import com.novoda.staticanalysis.internal.CollectViolationsTask
+import com.novoda.staticanalysis.internal.Violations
+
+class CollectPmdViolationsTask extends CollectViolationsTask {
+
+    @Override
+    void collectViolations(File xmlReportFile, File htmlReportFile, Violations violations) {
+        PmdViolationsEvaluator evaluator = new PmdViolationsEvaluator(xmlReportFile)
+        int errors = 0, warnings = 0
```
mr-archano: LoL, I tried.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/GenerateFindBugsHtmlReport.groovy
```diff
@@ -0,0 +1,19 @@
+package com.novoda.staticanalysis.internal.findbugs
+
+import org.gradle.api.tasks.JavaExec
+
+class GenerateFindBugsHtmlReport extends JavaExec {
```
ouchadam: 👍

ouchadam:
nice set of tests for each plugin, LGTM ! 👍

PR #32, opened 2017-04-14
Introduce dependencies plugin

mr-archano:
## Scope of the PR
We want to streamline the definition of dependencies in the project. We usually extract a [gradle script](https://github.com/mr-archano/Playground/blob/develop/gradle/dependencies.gradle) collecting all the dependencies as nested maps we can use in every sub project.
This approach, altho helps to tidy up the buildscripts across modules has a big disadvantage: there is no autocompletion provided by the IDE because the handles of each dependency are added dynamically to the properties extension of the project the script is applied to.

The solution proposed here is to introduce a custom plugin that simply exposes the centralised list of dependencies via project extensions:

- `libraries`: an instance of the `Libraries` class exposing a public property for each of the compile dependencies used in the sub-projects
- `gradlePlugins`: an instance of the `GradlePlugins` class exposing a public property for each of the gradle plugins used in the sub-projects

All you need to do is to apply the plugin to the root project and then access those project extensions, eg:
```
classpath gradlePlugins.bintrayRelease
```
or
```
testCompile libraries.junit
```

### DIFF
--- a/buildSrc/src/main/groovy/GradlePlugins.groovy
+++ b/buildSrc/src/main/groovy/GradlePlugins.groovy
@@ -1,4 +1,4 @@
-public class GradlePlugins {
+class GradlePlugins {
     final bintrayRelease = 'com.novoda:bintray-release:0.4.0'
     final buildProperties = 'com.novoda:gradle-build-properties-plugin:0.2'
     final gradleGit = 'org.ajoberstar:gradle-git:1.6.0'
--- a/buildSrc/src/main/groovy/GradlePlugins.groovy
+++ b/buildSrc/src/main/groovy/GradlePlugins.groovy
@@ -1,0 +1,5 @@
+public class GradlePlugins {
+    public final String bintrayRelease = 'com.novoda:bintray-release:0.4.0'
+    public final String buildProperties = 'com.novoda:gradle-build-properties-plugin:0.2'
+    public final String gradleGit = 'org.ajoberstar:gradle-git:1.6.0'
+}
--- a/buildSrc/src/main/groovy/GradlePlugins.groovy
+++ b/buildSrc/src/main/groovy/GradlePlugins.groovy
@@ -1,5 +1,5 @@
 public class GradlePlugins {
-    public final String bintrayRelease = 'com.novoda:bintray-release:0.4.0'
-    public final String buildProperties = 'com.novoda:gradle-build-properties-plugin:0.2'
-    public final String gradleGit = 'org.ajoberstar:gradle-git:1.6.0'
+    final bintrayRelease = 'com.novoda:bintray-release:0.4.0'
+    final buildProperties = 'com.novoda:gradle-build-properties-plugin:0.2'
+    final gradleGit = 'org.ajoberstar:gradle-git:1.6.0'
 }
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/groovy/Libraries.groovy
```diff
@@ -0,0 +1,27 @@
+import org.gradle.api.Project
+import org.gradle.api.artifacts.Dependency
+
+public class Libraries {
+    public final String junit = 'junit:junit:4.12'
```
tasomaniac: I'm just comparing this with the mentioned map we have. 
I only think that the map looks much cleaner and easier to manage. Here it is much verbose because it is Java. 

I was thinking if we can achieve what you want only with this: 

```
final junit = 'junit:junit:4.12'
```

without the types and public.

mr-archano: Yeah, good point. I wasn't too bothered of having something a little more verbose here, but I see your point. Thanks to groovy we can streamline this and still reap the benefits of using a class.

mr-archano:
@tasomaniac I applied your suggestion, it looks nicer than before: great idea 👍

plugin/build.gradle
```diff
@@ -23,13 +20,13 @@ sourceSets {
...
-    testCompile gradleTestKit()
-    testCompile 'junit:junit:4.12'
-    testCompile 'com.google.truth:truth:0.30'
-    testCompile 'com.google.guava:guava:19.0'
-    testCompile 'com.google.code.findbugs:jsr305:3.0.0'
+    testCompile libraries.gradleTestKit
+    testCompile libraries.junit
+    testCompile libraries.truth
+    testCompile libraries.guava
+    testCompile libraries.findbugs.annotations
```
ataulm: how comes this one would be like this vs. just `libraries.findbugs`?

I could see it for android support libs where there's several that fit under a single group but didn't follow the reasoning here.

mr-archano: I took the opportunity to stretch a little this trivial example introducing things we have already seen in another real life projects - just to get a feeling of how this will scale up.
So don't mind too much the ~nonsensical name here 😅

buildSrc/src/main/groovy/GradlePlugins.groovy
```diff
@@ -0,0 +1,5 @@
+class GradlePlugins {
```
devisnik: The constants should be accessible as `GradlePlugins.bintrayRelease` etc in all the build scripts. I suppose the IDE wasn't auto-completing on those?

buildSrc/src/main/groovy/Libraries.groovy
```diff
@@ -0,0 +1,27 @@
...
+
+class Libraries {
+    final junit = 'junit:junit:4.12'
+    final truth = 'com.google.truth:truth:0.30'
+    final guava = 'com.google.guava:guava:19.0'
+    final findbugs = new Findbugs()
+
+    private final Project project
+
+    public Libraries(Project project) {
```
devisnik: `public` shouldn't be needed here. 😄

buildSrc/src/main/groovy/GradlePlugins.groovy
mr-archano: Exactly @devisnik. This is just a cosmetic addition to have `gradlePlugins` vs `GradlePlugins` in our buildscripts.

buildSrc/src/main/groovy/Libraries.groovy
mr-archano: oh groovy... you gotta love it! (will remove)

Issue #36, opened 2017-10-15
Add support for Android Lint


rock3r:
The aim would be to have all the static analysis covered by this plugin in one go (obviously this only applies to Android modules)

mr-archano:
Support for Android lint should be straightforward, but we would need to define what concurs as warning and what as error (from the plugin point of view).

rock3r:
Addressed in #62

Issue #35, opened 2017-06-13
Add support for Detekt for Kotlin projects


rock3r:
Detekt is a static analysis tool for Kotlin. Since Kotlin adoption is growing fast, and the classic trifecta of static analysis plugins don't work with it, it would be nice to have some tool that covers Kotlin code too.
See https://github.com/arturbosch/detekt/

mr-archano:
Addressed in #63

PR #38, opened 2017-12-18
Start using Kotlin

rock3r:
This is the first step in Kotlinifying the plugin business logic, with the aim of getting rid of Groovy's black magic and hopefully making it more extensible and maintainable.

It's still a WIP and there's a lot to do, but the tests are still green, so that's something. This goes against a feature branch where all WIP PRs will end up, so that it is more manageable to review while it's still not done.

Pending TODOs from this PR:
 * Convention Mappings aren't seemingly accessible from Kotlin (they involve too much of Groovy's random magic), so the `ReportUrlRenderer` is currently injected directly. Will be addressed later on (or dropped — it's an internal API that we should likely not be using anyway)


mr-archano:
> Convention Mappings [...]  it's an internal API that we should likely not be using anyway

Convention maps aren't really Groovy related, but part of the Gradle architecture and AFAIK not a private API either (an interesting article from few years ago on the matter is here: https://dzone.com/articles/gradle-plugin-conventions). They are needed to provide lazy loading of configurations and their defaults in the plugin architecture.

It's also worth noting this plugin behaves as meta-plugin, that is a wrapper around a set of built-in plugins (Checkstyle, PMD, FindBugs) and follows what those plugins do/need in order to augment their API instead of provide limitations. I genuinely don't see what's the value of porting the whole plugin to Kotlin if we don't gain anything from it (removing Groovy code is not a valid motivation alone IMO).

What's the issue we want to solve by this migration?

rock3r:
@mr-archano `conventionMappings` are considered private API as much as I could see from Gradle people posting on the Gradle forum. Also the [Gradle 4.0 release notes](https://docs.gradle.org/4.0/release-notes.html) read:

> In the past, many plugin developers were forced to solve evaluation order problems by using the concept of convention mapping, an internal API in Gradle subject to change.

So it is definitely a private API. We should be using `PropertyState` instead: https://docs.gradle.org/4.0/release-notes.html#public-type-for-representing-lazily-evaluated-properties

As for the why converting to Kotlin, the reasoning is that it should be easier to code and maintain (less random Groovy magic happening, possibly easier to extend too and for others to contribute to), and we might get some minor perf gains too — although that is not the main point.

I personally want to contribute to this project — see #35 and #36 for — but unfortunately I cannot wrap my head around some of the stuff in the code because of the massive baggage of prior knowledge it requires of both Gradle and Groovy. It is my hope that using Kotlin would take away most of the implicit knowledge required to understand this codebase. It's not that it's not clear in how it works in broad strokes, or it's badly architected. Far from it, it's the best codebase I've seen so far in a Gradle plugin! Unfortunately, the Gradle APIs are not exactly easy to understand and master, and the laxness of Groovy on top of it makes it a daunting task. Having no help whatsoever from the compiler and the IDE is just the cherry on the top :( 

It remains that it's very complicated to work on and contribute to. I'm hoping at the very least to gain enough understanding of this project to be able to contribute to it. 

That said, I closed this PR because I am going to give up on the conversion. The heavily dynamic and strongly mutability-oriented nature of the code makes it tough to gradually port over to a more strict language such as Kotlin. The only possibility would be to have a full rewrite in Kotlin, but that is not worth it in my opinion.

mr-archano:
> So it is definitely a private API. We should be using PropertyState instead: https://docs.gradle.org/4.0/release-notes.html#public-type-for-representing-lazily-evaluated-properties

TIL! I guess I missed that big announcement then 😅 Thanks @rock3r I will look into that, sounds very promising indeed.
The reason why I kept working in Groovy over Java or Kotlin was mainly because I could reference the internal implementation of the configured plugins and borrow snippets of code to achieve my goal.
I understand why moving away from Groovy can make any contributor's life easier, so I will spare some time to investigate this migration/rewrite a little further.

I'll keep you posted here 🎅

rock3r:
No worries, and thanks again :) I did stop trying to convert the code myself, but would be very happy to help if we could do this together. In the meantime, happy holidays :)

PR #45, opened 2018-01-22
Make evaluation pass scriptable

mr-archano:
The evaluation of the violations at the moment is brought by the `EvaluateViolationsTask` but doesn't allow to do anything else but breaking the build and logging the outcome of the evaluation. We have already faced the need of being able to plug-in a different type of reporting for instance, but the current architecture is not particularly friendly in that regard.

The action performed in `EvaluateViolationsTask` has been extracted and modelled as `ViolationsEvaluator`. We let the plugin extension provide a way of (lazily) create such evaluator for the `EvaluateViolationsTask`.
By default the extension will provide an instance of `DefaultViolationsEvaluator`, but users can now also script their own evaluation phase as follows:
```
staticAnalysis {
   ...
   evaluator { Set<Violations> allViolations ->
       // add your evaluation logic here
   }
}
```

Defining a custom evaluator this way will override the default one.


mr-archano:
As briefly mentioned IRL I am more and more convinced that having the ability to script the evaluation phase would bring the most value. At this point though I would seriously consider to treat the thresholds only as configuration detail of the default evaluator, as another implementation could live completely outside the notion of thresholds. Thoughts, @tobiasheine, @devisnik?

plugin/src/main/groovy/com/novoda/staticanalysis/DefaultViolationsEvaluator.groovy
```diff
@@ -0,0 +1,45 @@
...
+
+    @Override
+    void evaluate(Set<Violations> allViolations) {
+        Map<String, Integer> total = [errors: 0, warnings: 0]
+        String fullMessage = '\n'
+        allViolations.each { Violations violations ->
+            int errors = violations.errors
+            int warnings = violations.warnings
+            if (errors > 0 || warnings > 0) {
+                fullMessage += "> ${getViolationsMessage(violations, reportUrlRenderer)}\n"
```
tasomaniac: Not really important but I was thinking if you separate this `fullMessage` and `total` construction in 2 different loops, it can be simplified a lot. 

```
allViolations.each { violations ->
  fullMessage += ${getViolationsMessage(violations, reportUrlRenderer)} // method adds \n and initial char if necessary
}
total['errors'] = allViolations.collect { it.errors }.sum()
total['warnings'] = allViolations.collect { it.warnings }.sum()
```

plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisExtension.groovy
```diff
@@ -60,4 +69,26 @@ class StaticAnalysisExtension {
...
+    ViolationsEvaluator getEvaluator() {
+        createEvaluator()
+    }
+
+    Logger getLogger() {
+        project.logger
+    }
+
+    void evaluator(Action<? super Set<Violations>> action) {
+        createEvaluator = {
```
tasomaniac: I don't understand why you need an extra `Closure` here? Is creating the internal evaluator really need to be lazy? Here, creating new ViolationsEvaluator and executing given action later with `evaluate` method already provides laziness, isn't it?

mr-archano: I suppose that was because of part of the `Input` needed to be accessed lazily (ie: thresholds after configuration). In general I believe this will allow us to avoid to wonder about similar issues in the action provided, don't you think?

plugin/src/main/groovy/com/novoda/staticanalysis/DefaultViolationsEvaluator.groovy
mr-archano: I wasn't aware of `.sum()`. #TIL

plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisExtension.groovy
mr-archano: ok, discussed IRL: the question was more around the need of having a closure to create the evaluator. The reason is that this way we can pass the closure (ie: factory) to the evaluation task instead of passing around the evaluator itself. Any thoughts @devisnik?

plugin/src/main/groovy/com/novoda/staticanalysis/DefaultViolationsEvaluator.groovy
tasomaniac: I'm not sure if it will work. :) Please double check. Looks cleaner.

plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisExtension.groovy
mr-archano: I just confirmed that the lazy creation of the evaluator will allow us to access the correct configured properties from the plugin extension(eg: `reportUrlRenderer`)

PR #46, opened 2018-01-23
PT-494: Support detekt in plain kotlin projects

tobiasheine:
### Description
[PT-494](https://novoda.atlassian.net/browse/PT-494)

This is the first PR to support [Detekt](https://github.com/arturbosch/detekt#gradleplugin), which is a static code analysis tool for Kotlin.

In this step we just support plain Kotlin projects. We therefore decided to target a feature branch instead of `develop`.

Also we expect the user to manually add the detekt plugin to its project:

```code
plugins {
    id "io.gitlab.arturbosch.detekt" version "1.0.0.RC6-2"
}
```

Additionally we expect the user to configure `output` and `config` for the detekt plugin:

```code
staticAnalysis {
    penalty {
        maxWarnings = 0
        maxErrors = 0
    }
    
    detekt { 
        profile('main') { 
          config = "detekt.yml" 
          output = "build/reports"
        }
    }
}
```

A default configuration can be found [here](https://github.com/arturbosch/detekt/blob/master/detekt-cli/src/main/resources/default-detekt-config.yml).

### Considerations

- extracted common `Configurator` interface which can be used by third party tools like detekt and android lint which don't extend from `org.gradle.api.plugins.quality.CodeQualityExtension`
- introduced `CollectDetektViolationsTask` which is a copy of `CollectCheckstyleViolationsTask` since the format is the same (thx @rock3r for point that out 👍 )
- introduced `DetektConfigurator` which configures `detekt` and collects the violations which are then considered by our threshold mechanism

### Tests
- added unit tests that verify `warnings` and `errors` are considered for threshold mechanism
- added unit tests that verify `warnings` and `errors` are ignored in case detekt is not conigured

Paired with @tasomaniac and @mr-archano on parts of it


plugin/src/test/groovy/com/novoda/test/TestProject.groovy
```diff
@@ -109,10 +120,18 @@ ${project.additionalConfiguration}
         projectDir.deleteDir()
     }
 
+    String projectDir(){
```
tobiasheine: I need to expose this to configure the `output` directory of the detekt plugin in the `DetektIntegrationTest `.

plugin/src/test/groovy/com/novoda/staticanalysis/internal/detekt/DetektIntegrationTest.groovy
```diff
@@ -0,0 +1,191 @@
...
+                    maxWarnings = 0
+                    maxErrors = 0
+                }''')
+
+        def detektConfiguration = """
+        detekt { 
+            profile('main') { 
+                config = "${Fixtures.Detekt.RULES}" 
+                output = "${testProject.projectDir()}/build/reports"
+                input = "${Fixtures.Detekt.SOURCES_WITH_WARNINGS}"
```
tobiasheine: The `input` just needs to be configured for the tests. Probably `detekt` doesn't pick up the changed source sets. In a example project it was not needed.

rock3r:
Would it be possible to automatically add the Detekt plugin in the future, so it can be configured to be a specific version number like the other plugins we support?

plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
```diff
@@ -0,0 +1,80 @@
...
+            if (project.tasks.findByName('detektCheck')) {
+                configureToolTask()
+            }
+        }
+    }
+
+    private void configureToolTask() {
+        def detektTask = project.tasks.findByName('detektCheck')
+        // run detekt as part of check
+        project.tasks.findByName('check').dependsOn(detektTask)
```
tasomaniac: Optional: Here is a more groovy way of accessing tasks. `project.tasks['check']`

plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
```diff
@@ -0,0 +1,80 @@
...
+
+        // evaluate violations after detekt
+        detektTask.group = 'verification'
+        def collectViolations = createCollectViolationsTask(violations)
+        evaluateViolations.dependsOn collectViolations
+        collectViolations.dependsOn detektTask
+    }
+
+    private CollectDetektViolationsTask createCollectViolationsTask(Violations violations) {
+        def reportFilePath = "${project.extensions.findByName('detekt').systemOrDefaultProfile().output}/detekt-checkstyle.xml"
```
tasomaniac: `systemOrDefaultProfile` seems very specific. I looked at the source code and saw that they have `profiles` set. And this method is just searching for 1 specific profile in that set. 

What about looping over all possible profiles? Something like: 

```
def detekt = project.extensions['detekt']
detekt.profiles.each { profile ->
  def reportFile = new File(profile.output, "detekt-checkstyle.xml")
  if (reportFile.exists) {
    // create tasks
  }
}
```

plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
tasomaniac: Optional: Here is a more groovy way of accessing tasks. `project.tasks['check']`

tobiasheine:
@rock3r we were planning to do that, but failed. I created a JIRA [task](https://novoda.atlassian.net/browse/PT-496) to follow-up on this.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
```diff
@@ -0,0 +1,80 @@
...
+
+        project.tasks.create("collectDetektViolations", CollectDetektViolationsTask) { collectViolations ->
+            collectViolations.xmlReportFile = new File(reportFilePath)
```
tasomaniac: In case we have multiple profiles, can we have multiple `xmlReportFile`? 
I think `detekt` supports multiple profiles.

plugin/src/test/groovy/com/novoda/staticanalysis/internal/detekt/DetektIntegrationTest.groovy
tasomaniac: These tests are like users of the plugin. Should we have this comment here in the tests? 
We spent couple of hours on this. Can be useful for future devs.

plugin/src/test/groovy/com/novoda/test/TestProjectRule.groovy
```diff
@@ -19,6 +19,10 @@ final class TestProjectRule<T extends TestProject> implements TestRule {
         new TestProjectRule({ new TestAndroidProject() }, { String name -> "project.android.sourceSets.$name" }, 'Android project')
     }
 
+    static TestProjectRule<TestJavaProject> forKotlinProject() {
```
tasomaniac: The return type here is `TestjavaProject`. Should it be `TestKotlinProject`?

rock3r:
I will do a review shortly as well, I'm SO looking forward to this :)

plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
tobiasheine: But i's `private`. Can we still access it?

tobiasheine: Doesn't look like `Could not get unknown property 'profiles' for DetektExtension` :(

tasomaniac: That's too bad. :( I wonder what happens when the user has different profiles set-up?
Anyways, this will cover the regular usage. Forget about my comment. 😄

plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
tobiasheine: Good point. Will add a JIRA task to follow-up on that 👍

tasomaniac: Since `profiles` is private, we cannot have multiple reports, right?

tobiasheine: Imho it's possible. The extension has a method to add profiles `fun profile(name: String, configuration: Action<in ProfileExtension>)`

```code
detekt { 
        profile('foo') { 
            output = "foo/reports"
        }

       profile('bar') { 
            output = "bar/reports"
        }
    }
```

tasomaniac: it is possible but how do you access them since `profiles` is private. I'm not sure if there is another way.

tobiasheine: Well, that's the challenge ^^

plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
rock3r: Maybe we should open an issue or a PR on Detekt for this? Looks like a reasonable tooling use case

plugin/src/test/groovy/com/novoda/staticanalysis/internal/detekt/DetektIntegrationTest.groovy
```diff
@@ -0,0 +1,194 @@
...
+                }''')
+                .build('check')
+
+        TestProject.Result result = testProject
+
+        assertThat(result.logs).doesNotContainDetektViolations()
+    }
+
+    @Test
+    void shouldNotFailWhenWarningsAreBelowThreshold() throws Exception {
```
rock3r: Nit: they're technically `within`, not `below` thresholds :)

plugin/src/test/groovy/com/novoda/staticanalysis/internal/detekt/DetektIntegrationTest.groovy
```diff
@@ -0,0 +1,194 @@
...
+        testProject = testProject.withToolsConfig(detektConfiguration)
+
+        TestProject.Result result = testProject
+                .build('check')
+
+        assertThat(result.logs).doesNotContainLimitExceeded()
+        assertThat(result.logs).doesNotContainDetektViolations()
+    }
+}
+
```
rock3r: Extra whiteline

plugin/src/test/groovy/com/novoda/test/Fixtures.groovy
```diff
@@ -44,4 +44,10 @@ public final class Fixtures {
         public static final File SAMPLE_REPORT = new File(REPORTS_DIR, 'findbugs/reports/sample.xml')
     }
 
+    final static class Detekt{
```
rock3r: `cmd-alt-l` plz

plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
tobiasheine: Good point, will add this to a ticket.

tobiasheine:
@rock3r comments are applied 👍

plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
tasomaniac: I opened a ticket in Detekt https://github.com/arturbosch/detekt/issues/705

PR #47, opened 2018-01-25
[PT-498]: Expect the user to apply detekt manually

tobiasheine:
[PT-498](https://novoda.atlassian.net/browse/PT-498)
### Description
We agreed [here](https://novoda.atlassian.net/browse/PT-496) to expect the user to apply the detekt plugin manually.

This PR will make the build fail and give the user a hint to apply the detekt plugin manually in case it's configured but not applied.

### Details
- throw an exception in case the detekt plugin is configured but not applied

### Tests
- changed the `TestKotlinProject` to not contain the detekt plugin per default
- apply the plugin in existing tests 
- add a test which verifies we fail gracefully in case the detekt plugin is configured but not applied

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
@@ -34,8 +34,6 @@ class DetektConfigurator implements Configurator {
                 return
             }
 
-            project.apply plugin: 'io.gitlab.arturbosch.detekt'
-
             project.extensions.findByName('detekt').with {
                 // apply configuration closure to detekt extension
                 config.delegate = it
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
@@ -36,7 +36,7 @@ class DetektConfigurator implements Configurator {
             }
 
             if (!project.tasks.findByName('detektCheck')) {
-                throw new GradleException('The Detekt plugin is configured but not applied. Please apply the plugin in your build script.')
+                throw new GradleException('The detekt plugin is configured but not applied. Please apply the plugin in your build script.')
             }
 
             project.extensions.findByName('detekt').with {
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
@@ -10,6 +10,9 @@ import org.gradle.api.Task
 
 class DetektConfigurator implements Configurator {
 
+    private static final String DETEKT_PLUGIN = 'io.gitlab.arturbosch.detekt'
+    private static final String DETEKT_NOT_APPLIED = 'The detekt plugin is configured but not applied. Please apply the plugin in your build script.'
+
     private final Project project
     private final Violations violations
     private final Task evaluateViolations
@@ -35,8 +38,8 @@ class DetektConfigurator implements Configurator {
                 return
             }
 
-            if (!project.tasks.findByName('detektCheck')) {
-                throw new GradleException('The detekt plugin is configured but not applied. Please apply the plugin in your build script.')
+            if (!project.plugins.hasPlugin(DETEKT_PLUGIN)) {
+                throw new GradleException(DETEKT_NOT_APPLIED)
             }
 
             project.extensions.findByName('detekt').with {
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
@@ -3,6 +3,7 @@ package com.novoda.staticanalysis.internal.detekt
 import com.novoda.staticanalysis.StaticAnalysisExtension
 import com.novoda.staticanalysis.internal.Configurator
 import com.novoda.staticanalysis.internal.Violations
+import org.gradle.api.GradleException
 import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.Project
 import org.gradle.api.Task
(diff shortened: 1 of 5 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
```diff
@@ -34,17 +35,17 @@ class DetektConfigurator implements Configurator {
                 return
             }
 
-            project.apply plugin: 'io.gitlab.arturbosch.detekt'
+            if (!project.tasks.findByName('detektCheck')) {
```
tasomaniac: What about `project.plugins.hasPlugin('plugin name')`?

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
```diff
@@ -34,7 +34,7 @@ ${project.additionalConfiguration}
                 .withPluginClasspath()
                 .forwardStdOutput(new OutputStreamWriter(System.out))
                 .forwardStdError(new OutputStreamWriter(System.out))
-        withPlugins('com.novoda.static-analysis')
+        withPlugin('com.novoda.static-analysis', null)
```
tasomaniac: Do we need this `null` here?

plugin/src/test/groovy/com/novoda/staticanalysis/internal/detekt/DetektIntegrationTest.groovy
```diff
@@ -190,4 +195,29 @@ class DetektIntegrationTest {
...
+        def detektConfiguration = """
+        detekt { 
+            profile('main') { 
+                config = "${Fixtures.Detekt.RULES}" 
+                output = "${testProject.projectDir()}/build/reports"
+            }
+        }
+        """
+
+        testProject = testProject.withToolsConfig(detektConfiguration)
```
mr-archano: more of a `nit:` but why extracting all these loval vars that bloat the test body? have you considered to extact a couple of methods in this class that would help to reuse the same setup across tests? (see `CheckstyleIntegrationTest` or similar when it comes to tool configuration)

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
```diff
@@ -129,7 +133,7 @@ ${project.additionalConfiguration}
     }
 
     protected static String formatPlugins(TestProject project) {
-        "${project.plugins.join(',\n')}"
```
mr-archano: was this working at all before? 💃

plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
```diff
@@ -3,12 +3,16 @@ package com.novoda.staticanalysis.internal.detekt
...
 import com.novoda.staticanalysis.internal.Violations
+import org.gradle.api.GradleException
 import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.Project
 import org.gradle.api.Task
 
 class DetektConfigurator implements Configurator {
 
+    private static final String DETEKT_PLUGIN = 'io.gitlab.arturbosch.detekt'
+    private static final String DETEKT_NOT_APPLIED = 'The detekt plugin is configured but not applied. Please apply the plugin in your build script.'
```
rock3r: Maybe some more informations would help users. At least a "refer to the Detekt documentation at https://github.com/arturbosch/detekt for further details"

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
```diff
@@ -84,15 +84,19 @@ ${project.additionalConfiguration}
...
     }
 
-    public T withPlugins(String... plugins) {
-        def formattedPlugins = plugins.collect { plugin ->
-            "id '$plugin'"
-        }.asList()
-
-        this.plugins.addAll(formattedPlugins)
+    public T withPlugin(String plugin, String version = null) {
+        this.plugins.add("id \"$plugin\" ${optionalVersionFrom(version)}")
```
mr-archano: `nit:` you can avoid to escape the double quotes if you use simple ones.

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
```diff
@@ -84,15 +84,19 @@ ${project.additionalConfiguration}
...
         return this
     }
 
+    private static String optionalVersionFrom(String version) {
```
mr-archano: maybe it's just me, but I would have inlined this as tri-state:
```
this.plugins.add("id '$plugin' ${version?"version '$version'" : ""}")
```

plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
```diff
@@ -34,17 +38,17 @@ class DetektConfigurator implements Configurator {
...
             project.extensions.findByName('detekt').with {
                 // apply configuration closure to detekt extension
                 config.delegate = it
                 config()
             }
 
-            if (project.tasks.findByName('detektCheck')) {
-                configureToolTask()
-            }
+            configureToolTask()
```
mr-archano: just looked at the implementation of `configureToolTask()`: does it mean that Detekt generates a single task per project? how do you setup separate source sets and filters on them?

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
tasomaniac: 👍 for single quotes

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
tasomaniac: tests were passing :)
I think it was working because we never put more than 1 plugin. 😄

tobiasheine: exactly ^^

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
tobiasheine: Way better 👍  wasn't aware this is possible in groovy

plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
tobiasheine: We rely on detekt to pick them up. In case the user doesn't provide any `input` detekt automatically looks up different modules.

If you want to have multiple source sets with different filters, you need to configure different `profiles`. As for now we can just access the `main` profile. @tasomaniac opened already an [issue](https://github.com/arturbosch/detekt/issues/705) on the detekt repo to expose other profiles as well.

tobiasheine: As we spoke IRL, @mr-archano is referring to our filter feature. I will add another subtask to evaluate whether we can support that.

tasomaniac: 👍

tasomaniac:
All comments addressed. Merging this then.

PR #48, opened 2018-01-26
[PT-495] Support android kotlin projects

tobiasheine:
[PT-495](https://novoda.atlassian.net/browse/PT-495)

### Description
Scope of this PR is to add to make sure the detekt integration works also with android kotlin projects. Turned out that this works out of the box, so this PR just modifies the tests to consider also android kotlin projects.

### Tests
- created `TestAndroidKotlinProject`
- changed `DetektIntegrationTest` to run all existing tests also against a `TestAndroidKotlinProject`

### DIFF
--- a/plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
+++ b/plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
@@ -1,0 +1,78 @@
+package com.novoda.test
+
+class TestAndroidKotlinProject extends TestProject<TestAndroidKotlinProject> {
+    private static final Closure<String> TEMPLATE = { TestAndroidKotlinProject project ->
+        """
+buildscript {
+    repositories { 
+        google()
+        jcenter()
+    }
+    dependencies {
+        classpath 'com.android.tools.build:gradle:3.0.1'  
+        classpath 'org.jetbrains.kotlin:kotlin-gradle-plugin:1.2.20'
+    }
+}
+plugins {
+    ${formatPlugins(project)}
+}
+repositories { 
+    google()
+    jcenter()
+}
+apply plugin: 'com.android.library'  
+apply plugin: 'kotlin-android'
+
+android {
+    compileSdkVersion 27
+    buildToolsVersion '27.0.0'
+
+    defaultConfig {
+        minSdkVersion 16
+        targetSdkVersion 27
+        versionCode 1
+        versionName '1.0'
+    }
+    sourceSets {
+        ${formatSourceSets(project)}
+    }
+    ${project.additionalAndroidConfig}
+}
+${formatExtension(project)}
+"""
+    }
+
+    private String additionalAndroidConfig = ''
+
+    TestAndroidKotlinProject() {
+        super(TEMPLATE)
+        File localProperties = Fixtures.LOCAL_PROPERTIES
+        if (localProperties.exists()) {
+            withFile(localProperties, 'local.properties')
+        }
+    }
+
+    private static String formatSourceSets(TestProject project) {
+        project.sourceSets
+                .entrySet()
+                .collect { Map.Entry<String, List<String>> entry ->
+            """$entry.key {
+            manifest.srcFile '${Fixtures.ANDROID_MANIFEST}'
+            java {
+                ${entry.value.collect { "srcDir '$it'" }.join('\n\t\t\t\t')}
+            }
+        }"""
+        }
+        .join('\n\t\t')
+    }
+
+    @Override
+    List<String> defaultArguments() {
+        ['-x', 'lint'] + super.defaultArguments()
+    }
+
+    TestAndroidKotlinProject withAdditionalAndroidConfig(String additionalAndroidConfig) {
+        this.additionalAndroidConfig = additionalAndroidConfig
+        return this
+    }
+}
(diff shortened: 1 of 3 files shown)


plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
```diff
@@ -0,0 +1,78 @@
+package com.novoda.test
```
tobiasheine: This is basically a copy of `TestAndroidProject`. Changes are marked.

plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
```diff
@@ -0,0 +1,78 @@
...
+    private static final Closure<String> TEMPLATE = { TestAndroidKotlinProject project ->
+        """
+buildscript {
+    repositories { 
+        google()
+        jcenter()
+    }
+    dependencies {
+        classpath 'com.android.tools.build:gradle:3.0.1'  
+        classpath 'org.jetbrains.kotlin:kotlin-gradle-plugin:1.2.20'
```
tobiasheine: This has been added

plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
```diff
@@ -0,0 +1,78 @@
...
+}
+plugins {
+    ${formatPlugins(project)}
+}
+repositories { 
+    google()
+    jcenter()
+}
+apply plugin: 'com.android.library'  
+apply plugin: 'kotlin-android'
```
tobiasheine: This has been added

plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    private static String formatSourceSets(TestProject project) {
+        project.sourceSets
+                .entrySet()
+                .collect { Map.Entry<String, List<String>> entry ->
+            """$entry.key {
+            manifest.srcFile '${Fixtures.ANDROID_MANIFEST}'
+            kotlin {
```
tobiasheine: This has been changed

plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
tasomaniac: to be honest you can also use the `plugins` and your `withPlugin` method to add these dynamically from test code. 
The plugin id is `org.jetbrains.kotlin.android`
I think also `com.android.library` works too.

plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
tasomaniac: well, if you add your sources to `java` folder, you don't need this in test setup.

plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
rock3r: I think you may have left two whitespaces at the end of the line before when adding it:

image

EDIT: actually, there seem to be a few :) Nothing serious. Just my nitpicky eye 💃

plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
tasomaniac: It is not too important but with the comments I put below, it is possible that these can be 1 single class with different plugin configuration. 

Then you can have `TestKotlinProject.forAndroid()` to configure the android plugin.

plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
takecare: 👁‍🗨

plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
tobiasheine: I actually tried that for the kotlin plugin, but it didn't work. I got this error `Extension with name 'android' does not exist`. 
And afaik this is because of the order. `com.android.library` needs to be applied first and I couldn't find that one [here](https://plugins.gradle.org/search?term=android+library&page=2).

plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
```diff
@@ -0,0 +1,78 @@
...
+            withFile(localProperties, 'local.properties')
+        }
+    }
+
+    private static String formatSourceSets(TestProject project) {
+        project.sourceSets
+                .entrySet()
+                .collect { Map.Entry<String, List<String>> entry ->
+            """$entry.key {
+            manifest.srcFile '${Fixtures.ANDROID_MANIFEST}'
```
takecare: maybe use double quotes just so there's more consistency? you can also just use `$x.y` here i believe

plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
tobiasheine: I tried that first but failed (see my comment below). Any help is appreciated since I also would like to get rid of this duplication.

plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
tasomaniac: This is printing the entire config. In the generated one we want single quotes of possible.

plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
tasomaniac: Just merged it. On a second thought, I don't think it is worth. Because that way, it will be more complicated.

PR #49, opened 2018-01-26
[PT-503] Detekt Integration Bugs

tobiasheine:
[PT-503](https://novoda.atlassian.net/browse/PT-503)

### Description
While testing the integration in the wild we found two bugs:
- the detekt reports are not picked up for evaluation
- the link to the html report is wrong

### Considerations
- generate path to report file by using `project.file()` with the relative path. This was working in the tests since we are using an absolute path
- make `htmlReportFile` configurable since detekt names the html report different than the xml report.

### Tests
- added a test to verify the link to the html report is correct

Paired with @tasomaniac


tasomaniac:
Merging since it resolves bugs.

Issue #43, opened 2018-01-20
Use file output to store violations per-tool


mr-archano:
Right now the plugin stores the violations in an in-memory `NamedDomainObjectContainer` holding an instance of `Violations` for each of the tool configured. Ideally each tool would instead produce an (xml?) report that can be used as input of the `EvaluateViolationsTask`. This would also help with the incremental builds, as an input/output file plays nicer with Gradle architecture.

tasomaniac:
You can define custom Objects to be input/output too. With `Property` and `Provider`s. (Are they Gradle 4+?) I think they also need to be Serializable then. The incremental build mechanism works like the same. Writing to a file may not be necessary.

mr-archano:
@tasomaniac good point, but then 2 things come to mind:
- 1. I don't see why we should loose retro-compatibility, even though Gradle 4.x is widely adopted already
- 2. I feel like having the output as persisted report can open possibilities we haven't fully figured out yet, even though we can always write the objects as files if we need to

Mmmm.

tasomaniac:
That may makes sense for other purposes of course. Since all the tools integrated already has output files, it could also be an overhead to have even more files. ☺

PR #51, opened 2018-01-30
[PT-511] Configure lint via SAP extension

tobiasheine:
[PT-511](https://novoda.atlassian.net/browse/PT-511)

### Description
Scope of this PR is to add basic support for [Android Lint](https://developer.android.com/studio/write/lint.html) to the SAP plugin.

This targets a feature branch.

### Considerations
- Added `CollectLintViolationsTask` which parses `lint-result.xml` and appends errors and warnings to the violations

- Added `LintConfigurator` which applies the `lintConfig` closure to Lint and orchestrates the tasks

### Tests
- Added a test for the `CollectLintViolationsTask` to verify we parse the `lint-result.xml` correctly
- Added the `LintIntegrationTest` which verifies warnings and errors are considered for the thresholds

Paired with @lgvalle on parts of it

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
@@ -48,7 +48,7 @@ class LintConfigurator implements Configurator {
 
     private CollectLintViolationsTask createCollectViolationsTask(Violations violations) {
         def outputFolder = new File(project.projectDir, 'build/reports')
-        project.tasks.create("collectLintViolations", CollectLintViolationsTask) { collectViolations ->
+        project.tasks.create('collectLintViolations', CollectLintViolationsTask) { collectViolations ->
             collectViolations.xmlReportFile = new File(outputFolder, 'lint-results.xml')
             collectViolations.htmlReportFile = new File(outputFolder, 'lint-results.html')
             collectViolations.violations = violations
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
@@ -50,8 +50,8 @@ class LintConfigurator implements Configurator {
     private CollectLintViolationsTask createCollectViolationsTask(Violations violations) {
         def outputFolder = new File(project.projectDir, 'build/reports')
         project.tasks.create("collectLintViolations", CollectLintViolationsTask) { collectViolations ->
-            collectViolations.xmlReportFile = new File(outputFolder, 'lint-report.xml')
-            collectViolations.htmlReportFile = new File(outputFolder, 'lint-report.html')
+            collectViolations.xmlReportFile = new File(outputFolder, 'lint-results.xml')
+            collectViolations.htmlReportFile = new File(outputFolder, 'lint-results.html')
             collectViolations.violations = violations
         }
     }
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
@@ -48,7 +48,6 @@ class LintConfigurator implements Configurator {
     }
 
     private CollectLintViolationsTask createCollectViolationsTask(Violations violations) {
-//        def outputFolder = project.file(project.extensions.findByName('android').lintOptions.xmlOutput.parent)
         def outputFolder = new File(project.projectDir, 'build/reports')
         project.tasks.create("collectLintViolations", CollectLintViolationsTask) { collectViolations ->
             collectViolations.xmlReportFile = new File(outputFolder, 'lint-report.xml')
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
@@ -66,7 +66,7 @@ class LintConfigurator implements Configurator {
     }
 
     private File getDefaultOutputFolder() {
-        new File(project.projectDir, 'build/reports')
+        new File(project.buildDir, 'reports')
     }
 
     private static boolean isAndroidProject(final Project project) {
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
@@ -48,7 +48,8 @@ class LintConfigurator implements Configurator {
     }
 
     private CollectLintViolationsTask createCollectViolationsTask(Violations violations) {
-        def outputFolder = project.file(project.extensions.findByName('android').lintOptions.xmlOutput.parent)
+//        def outputFolder = project.file(project.extensions.findByName('android').lintOptions.xmlOutput.parent)
+        def outputFolder = new File(project.projectDir, 'build/reports')
         project.tasks.create("collectLintViolations", CollectLintViolationsTask) { collectViolations ->
             collectViolations.xmlReportFile = new File(outputFolder, 'lint-report.xml')
             collectViolations.htmlReportFile = new File(outputFolder, 'lint-report.html')
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
@@ -54,12 +54,19 @@ class LintConfigurator implements Configurator {
     }
 
     private CollectLintViolationsTask createCollectViolationsTask(Violations violations) {
-        def outputFolder = new File(project.projectDir, 'build/reports')
         project.tasks.create('collectLintViolations', CollectLintViolationsTask) { collectViolations ->
-            collectViolations.xmlReportFile = new File(outputFolder, 'lint-results.xml')
-            collectViolations.htmlReportFile = new File(outputFolder, 'lint-results.html')
+            collectViolations.xmlReportFile = xmlOutputFile()
+            collectViolations.htmlReportFile = new File(defaultOutputFolder(), 'lint-results.html')
             collectViolations.violations = violations
         }
+    }
+
+    private File xmlOutputFile() {
+        project.android.lintOptions.xmlOutput ?: new File(defaultOutputFolder(), 'lint-results.xml')
+    }
+
+    private File defaultOutputFolder() {
+        new File(project.projectDir, 'build/reports')
     }
 
     private static boolean isAndroidProject(final Project project) {
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
@@ -41,6 +41,7 @@ class LintConfigurator implements Configurator {
         project.android.lintOptions(config)
         project.android.lintOptions {
             xmlReport = true
+            htmlReport = true
             abortOnError false
         }
     }
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
@@ -34,7 +34,7 @@ class LintConfigurator implements Configurator {
                 return
             }
 
-            project.extensions.findByName('android').lintOptions(config)
+            project.android.lintOptions(config)
 
             configureToolTask()
         }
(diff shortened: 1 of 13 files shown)


plugin/src/test/groovy/com/novoda/staticanalysis/internal/detekt/DetektIntegrationTest.groovy
```diff
@@ -15,11 +13,6 @@ import static com.novoda.test.LogsSubject.assertThat
 @RunWith(Parameterized.class)
 class DetektIntegrationTest {
 
-    @ClassRule
```
tobiasheine: Just some boy scouting, we're not using the artifact in this test.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
```diff
@@ -0,0 +1,64 @@
...
+
+    private void configureToolTask() {
+        // evaluate violations after lint
+        def collectViolations = createCollectViolationsTask(violations)
+        evaluateViolations.dependsOn collectViolations
+        collectViolations.dependsOn project.tasks['lint']
+    }
+
+    private CollectLintViolationsTask createCollectViolationsTask(Violations violations) {
+        def outputFolder = new File(project.projectDir, 'build/reports')
```
tasomaniac: We shouldn't hardcode `build/reports`, android team may change the location. I wonder if the `lint` task has output?

plugin/src/test/groovy/com/novoda/test/TestAndroidProject.groovy
```diff
@@ -63,11 +63,6 @@ ${formatExtension(project)}
         .join('\n\t\t')
     }
 
-    @Override
```
tobiasheine: @mr-archano what was the reasons for this? Can we kill it? All tests are still passing...

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
```diff
@@ -0,0 +1,64 @@
...
+
+    @Override
+    void execute() {
+        project.extensions.findByType(StaticAnalysisExtension).ext."lintOptions" = { Closure config ->
+
+            if (!isAndroidProject(project)) {
+                return
+            }
+
+            project.extensions.findByName('android').lintOptions(config)
```
tasomaniac: I'm not sure if you like more groovy magic but I believe this can be replaced by 
`project.android.lintOptions(config)`

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
```diff
@@ -0,0 +1,64 @@
...
+class LintConfigurator implements Configurator {
+
+    private final Project project
+    private final Violations violations
+    private final Task evaluateViolations
+
+    static LintConfigurator create(Project project,
+                                   NamedDomainObjectContainer<Violations> violationsContainer,
+                                   Task evaluateViolations) {
+        Violations violations = violationsContainer.maybeCreate('Lint')
```
tasomaniac: I'm confused by this `maybeCreate('Lint')`
This means that the container is supposed to have a named closure called `Lint` but I couldn't find another usage. Not in the tests for example.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
```diff
@@ -0,0 +1,64 @@
...
+    private void configureToolTask() {
+        // evaluate violations after lint
+        def collectViolations = createCollectViolationsTask(violations)
+        evaluateViolations.dependsOn collectViolations
+        collectViolations.dependsOn project.tasks['lint']
+    }
+
+    private CollectLintViolationsTask createCollectViolationsTask(Violations violations) {
+        def outputFolder = new File(project.projectDir, 'build/reports')
+        project.tasks.create("collectLintViolations", CollectLintViolationsTask) { collectViolations ->
```
rock3r: Nit: use single quotes for `"collectLintViolations"` for consistency

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
tobiasheine: As follow-up it would be good to resolve this path dynamically. Otherwise this will break in case the user configures another `xmlOutput`

tobiasheine: Yep, I agree. Also the user can change the location.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
tobiasheine: I am also confused since I don't fully understand what's going in here. Tbh I just copy pasted it from the other configurators. From what I understand we pass the same `Violations` to each `Configurator`. But I don't understand how this relates to it.

plugin/src/test/groovy/com/novoda/staticanalysis/internal/lint/CollectLintViolationsTaskTest.groovy
```diff
@@ -0,0 +1,24 @@
...
+import static com.google.common.truth.Truth.assertThat
+
+class CollectLintViolationsTaskTest {
+
+    @Test
+    void shouldAddResultsToViolations() throws Exception {
+        Project project = ProjectBuilder.builder().build()
+        def task = project.task('collectLintViolationsTask', type: CollectLintViolationsTask)
+
+        Violations violations = new Violations("Android Lint")
```
rock3r: Nit: use single quotes for `"Android Lint"` for consistency

plugin/src/test/fixtures/sources/lint/warnings/Warning.java
```diff
@@ -0,0 +1,7 @@
+public class Warning {
+
+    public Warning() {
+        // assertions are ignored at runtime
```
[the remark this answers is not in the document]
tobiasheine: Nope, using assertions in code results in a `Warning` for lint. This is just a fixture that can be used for tests when provoking a lint warning.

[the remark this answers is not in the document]
rock3r: Oh I'm blind and I hadn't seen the path. My bad, sorry :)

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
tasomaniac: Maybe @mr-archano can enlighten us around this.

mr-archano: We just hold all the violations in the plugin extension to keep the output of each tool output in memory before the final evaluation (see [`StaticAnalysisExtension.allViolations`](https://github.com/novoda/gradle-static-analysis-plugin/blob/develop/plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisExtension.groovy#L72)).
Each tool is now responsible of creating the `Violations` entry in the container, that is then passed to the [`evaluateViolations`](https://github.com/novoda/gradle-static-analysis-plugin/blob/develop/plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisPlugin.groovy#L29) task.

As already mentioned IRL couple of times this is a pretty lame way of collecting all the info, and I already logged an issue to get rid of this in the next future: https://github.com/novoda/gradle-static-analysis-plugin/issues/43

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    void execute() {
+        project.extensions.findByType(StaticAnalysisExtension).ext."lintOptions" = { Closure config ->
+            if (!isAndroidProject(project)) {
+                return
+            }
+
+            project.android.lintOptions(config)
```
mr-archano: This seems to be missing the bit where we avoid lint to break the build (we want to be in charge of when doing that). I believe we should force `abortOnError false` here.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
mr-archano: we should fix this now, as we shouldn't hardcode anything that can be configured. The only constraint we actually have is that we need to get this information after lint is configured, in order to read the correct path to the xml report. We have two ways of doing this as I see it: we get it from `lintOptions` or from the `lint` task.
If we assume that the user will configure lint through our plugin - effectively done at L36 - then we should read the value from the lint extension right afterwards, right?

plugin/src/test/groovy/com/novoda/test/TestAndroidProject.groovy
mr-archano: the reason for this was to temporary disable lint as the default configuration breaks the build on error. I would have expected the other integration tests to fail if these lines are removed, mmm

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
tasomaniac: Also keep in mind that `lint` is run for all variants. Maybe we should have `includeVariants` here too. But in any case, should we also support multiple outputs?

tasomaniac: @mr-archano this PR is towards a feature branch. Not on develop. So I think we can just merge it without this? with a follow-up soon?

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
tasomaniac: Having a look at the whole options here: https://google.github.io/android-gradle-dsl/current/com.android.build.gradle.internal.dsl.LintOptions.html

I think we can also add `xmlReport true` to make sure that xml report is enabled.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
tobiasheine: @mr-archano just checked that, but the output path of `lintOptions` is null in case the user hasn't configured it. So we need to find another way. I agree with @tasomaniac to follow up on this as part of [this](https://novoda.atlassian.net/browse/PT-516) ticket.

tobiasheine: We agreed on for now to check whether the output path of `lintOptions` is configured or to fall back to the hardcoded one.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
tobiasheine: Same for the htmlReport I'd say

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
tobiasheine: @mr-archano done

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
```diff
@@ -53,14 +54,21 @@ class LintConfigurator implements Configurator {
...
         project.tasks.create('collectLintViolations', CollectLintViolationsTask) { collectViolations ->
-            collectViolations.xmlReportFile = new File(outputFolder, 'lint-results.xml')
-            collectViolations.htmlReportFile = new File(outputFolder, 'lint-results.html')
+            collectViolations.xmlReportFile = xmlOutputFile()
+            collectViolations.htmlReportFile = new File(defaultOutputFolder(), 'lint-results.html')
             collectViolations.violations = violations
         }
     }
 
+    private File xmlOutputFile() {
```
mr-archano: sorry to be a PITA, but this could be a getter `getXmlOutputFile()`, so you can use it simply at L58 as `xmlOutputFile` property

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
```diff
@@ -53,14 +54,21 @@ class LintConfigurator implements Configurator {
...
+        project.android.lintOptions.xmlOutput ?: new File(defaultOutputFolder(), 'lint-results.xml')
+    }
+
+    private File defaultOutputFolder() {
```
mr-archano: same here

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
```diff
@@ -53,14 +54,21 @@ class LintConfigurator implements Configurator {
...
+        new File(project.projectDir, 'build/reports')
```
mr-archano: you should never add `build/`, but rather use `project.buildDir`. this will become then:
```
new File(project.buildDir, 'reports')
```

PR #53, opened 2018-01-30
Tweak Detekt missing exception message

rock3r:
This PR tweaks the Detekt missing error message. Just adding a missing full stop and a new line for readability.

PR #54, opened 2018-01-30
Add Detekt docs and reorganise existing readme/docs

rock3r:
This PR:
 * Fixes the Apache 2.0 license setup, adding the `NOTICE` file and reformatting/fixing `LICENSE` (was `LICENSE.txt`)
 * Improves the readme file, trimming it down to the bare essentials to get people going without overwhelming them
 * Moves the advanced configuration docs to the `docs` folder
 * Adds detailed informations about all supported tools, in the `docs` folder
 * Many more small tweaks and grammar improvements


docs/supported-tools.md
```diff
@@ -0,0 +1,271 @@
...
+[Detekt](https://github.com/arturbosch/detekt) is a static analysis tool that looks for potential bugs and style violations in Kotlin code. It
+does not support Java. It can be used in both pure Kotlin, and Android Kotlin projects. It then only makes sense to have Detekt enabled if you
+have Kotlin code in your project. The plugin only runs Detekt on projects that contain the Kotlin or the Kotlin-Android plugin.
+
+### IMPORTANT: setup Detekt
+Unlike the other tools, the plugin **won't automatically add Detekt** to your project. If you forget to do it, the plugin will fail the build
+with an error like:
+
+```
+The detekt plugin is configured but not applied. Please apply the plugin in your build script.
```
tasomaniac: Do we really need to go into this much details? Above paragraph is I think really good but the duplication of the error message is redundant. I don't think this will be helpful to anybody.

docs/supported-tools.md
```diff
@@ -0,0 +1,271 @@
...
+For more information see https://github.com/arturbosch/detekt.
+```
+
+In order to use Detekt in your project, you need to:
+
+ 1. Add this statement to your root `build.gradle` project (change the version according to your needs):
```
tasomaniac: Actually this is not entirely true. In a simpler project where there is only 1 subproject, you can directly have this there without having anything in `root` at all.

docs/supported-tools.md
```diff
@@ -0,0 +1,271 @@
...
+### IMPORTANT: setup Detekt
+Unlike the other tools, the plugin **won't automatically add Detekt** to your project. If you forget to do it, the plugin will fail the build
+with an error like:
+
+```
+The detekt plugin is configured but not applied. Please apply the plugin in your build script.
+For more information see https://github.com/arturbosch/detekt.
+```
+
+In order to use Detekt in your project, you need to:
```
tasomaniac: What about having a link to official Detekt documentation? Users will definitely have to go there for details of the configuration anyways. So we would like them to go there and use their docs to learn how to add. And then use our docs to setup.

README.md
```diff
@@ -1,23 +1,43 @@
...
-- flexible, configurable penalty strategy for builds,
-- easy, Android-friendly integration for all static analysis,
-- convenient way of sharing same setup across different projects,
-- healthy, versionable and configurable defaults.
+- flexible, configurable penalty strategy for builds
+- easy, Android-friendly integration for all static analysis
+- convenient way of sharing same setup across different projects
+- healthy, versionable and configurable defaults
+
+### Supported static analysis tools
```
tasomaniac: `Supported tools` to have a shorter title?

README.md
```diff
@@ -1,23 +1,43 @@
...
+Support for additional tools is planned but not available yet:
 
-## Adding to your project
+ * [`KtLint`](https://github.com/shyiko/ktlint)
+ * [`Android Lint`](https://developer.android.com/studio/write/lint.html)
+
+Please note that the tools availability depends on the project the plugin is applied to. For more details please refer to the
+[supported tools](docs/supported-tools.md) page.
+
+### Out-of-the-box support for Android and Kotlin projects
```
tasomaniac: Title includes `kotlin` but the content does not mention anything kotlin related.

README.md
```diff
@@ -1,23 +1,43 @@
-# gradle-static-analysis-plugin
-[![](https://ci.novoda.com/buildStatus/icon?job=gradle-static-analysis-plugin)](https://ci.novoda.com/job/gradle-static-analysis-plugin/lastSuccessfulBuild/console) [![](https://img.shields.io/badge/License-Apache%202.0-lightgrey.svg)](LICENSE.txt) [![Bintray](https://api.bintray.com/packages/novoda/maven/gradle-static-analysis-plugin/images/download.svg) ](https://bintray.com/novoda/maven/gradle-static-analysis-plugin/_latestVersion)
+# Gradle static analysis plugin
```
tasomaniac: What about having a table of contents like section to have links to advanced configuration etc? Because otherwise, I'm afraid people will not see the inlined links.

rock3r: That wouldn't be a TOC of this document... besides there's extremely little content right now, I don't think it's easy to miss the links

README.md
rock3r: Same for me, doesn't really change much ¯\\\_(ツ)\_/¯

README.md
rock3r: Not much to mention besides that? I was not sure what to add. Suggestions welcome.

docs/supported-tools.md
rock3r: Reasoning: you get the error message and search for it. You get a hit, with the resolution steps. Everyone's happy.

README.md
tasomaniac: If we don't, maybe we can remove `Kotlin` from this header? Having support for `detekt` `ktlint` already indicates that we support those tools.

docs/supported-tools.md
rock3r: There is several already, specifically pointing to the relevant bits, where applicable. Our instructions on how to add are different enough from the official ones that they are conflicting, so I chose not to add a direct reference here (but there are in the next sections)

docs/supported-tools.md
rock3r: Well, not true, because Detekt expects to be applied to the root project, _and then_ to all relevant subprojects.

docs/supported-tools.md
tasomaniac: Then I would put a link to our docs in the error message. Rather than having it duplicated here. 
They would see the error and click the link, that's all. Don't have to search for something. 

Duplication is bad because the error may change and it is hard to maintain this copy to be the same.

docs/supported-tools.md
tasomaniac: My goal was to keep this docs brief. Considering we will have more tools, having too much information may hurt. 
"How to add `detekt`" is out of our concern, imho. We expect it to be added, that's all.

docs/supported-tools.md
tasomaniac: Either works. You don't have to have it in the root. If you have it in the root, then you don't need to duplicate the version. 

My goal around this comment actually was that I think we have too much information here.

README.md
rock3r: Well, does it imply it just as much as mentioning Android Lint implies support for Android projects? I don't _really_ see the problem here, to be honest.

docs/supported-tools.md
rock3r: That's a good idea, tomorrow I will change #53 for that to change the text here _and_ add a link to this bit.

docs/supported-tools.md
rock3r: I think it would be a grave omission not to have this information here — after all you suggest yourself at line 222 to link to this documentation in the error message.

When we'll add more tools, we'll begin splitting this out into separate docs. I'm already planning that, but it doesn't really make much sense _right now_ imho. I will do it when I add Lint.

docs/supported-tools.md
rock3r: >If you have it in the root, then you don't need to duplicate the version.
In the case you were talking about initially, there is only the root, so there is no duplication. In the "normal" case that this guide covers, there is no duplication either since the version is only written in one place, the root `build.gradle`.

>My goal around this comment actually was that I think we have too much information here.
Do you mean it's bad having the minimum instructions to make the project work? I mean, this is literally the minimum amount of steps required for the Detekt integration to work. If you meant that, as things are, adding Detekt support through SAP is a painful process that requires a lot of explanations to a new starters, I absolutely agree — I maintain that not applying the plugin ourselves makes this UX rather unsatisfactory and it is _not_ fine to leave things as they are in the long term

docs/supported-tools.md
```diff
@@ -0,0 +1,271 @@
...
+
+```gradle
+staticAnalysis {
+    penalty {
+        // ...
+    }
+}
+```
+
+## Checkstyle
```
mr-archano: @rock3r what about creating a separate md file for each tool?

rock3r: As mentioned above, was planning to do it when adding Lint. Can do in this PR if consensus is going that way.

README.md
tobiasheine: I agree with @tasomaniac, imho we could remove `Kotlin` from the headline. Additionally we could be a bit more explicit by mentioning `PMD`, `Findbugs` and `Checkstyle` as problematic when it comes to Android Projects.

docs/advanced-usage.md
```diff
@@ -0,0 +1,94 @@
...
+> Checkstyle rule violations were found (0 errors, 1 warnings). See the reports at:
+- http://ci.mycompany.com/job/myproject/ws/app/build/reports/checkstyle/main.html
+```
+
+And that will make them easier to follow them to the respective reports. More info on the topic can be found in the
+[`LogsExtension`](blob/master/plugin/src/main/groovy/com/novoda/staticanalysis/LogsExtension.groovy)
+Groovydocs.
+
+## Add exclusions with `exclude` filters
+You can specify custom patterns to exclude specific files from the static analysis. All you have to do is to specify `exclude`
```
tobiasheine: This doesn't work for `Detekt` as of now. Should we mention that?

docs/advanced-usage.md
```diff
@@ -0,0 +1,94 @@
...
+        exclude '**/*Test.java' // file pattern
+        exclude project.fileTree('src/test/java') // entire folder
+        exclude project.file('src/main/java/foo/bar/Constants.java') // specific file
+        exclude project.sourceSets.main.java.srcDirs // entire source set
+    }
+}
+```
+
+## Add exclusions with Android build variants
+Sometimes using `exclude` filters could be not enough. When using the plugin in an Android project you may want to consider
```
tobiasheine: Same as above

docs/supported-tools.md
```diff
@@ -0,0 +1,271 @@
...
+    ```
+ 2. Add this statement to each Kotlin project's `build.gradle`s:
+    ```gradle
+    plugins {
+        id 'io.gitlab.arturbosch.detekt'
+        // ...
+    }
+    ```
+
+### Configure Detekt
```
tobiasheine: Should we add that we suggest to disable thresholds in the `detekt` config file?

README.md
tasomaniac: We have a special paragraph to talk about `Android` because the default tools needs setup for Android and does not work out of the box. 
That's why this paragraph is really good. 
We don't have anything special around Kotlin, imo

docs/supported-tools.md
tasomaniac: Let me clarify :) What I mean is that our plugin does not care how Detekt plugin is added. We just expect it to be available. 
There are many ways to add a plugin. And can also change in the future. So that's why I thought we can leave it to the official docs. But it is not really that important.

docs/supported-tools.md
tasomaniac: Good point. Since we cannot automatically do this, it is really important to mention.

docs/supported-tools.md
rock3r: After the comment from @mr-archano below asking for a split I put some more thought into the split. I will do it, before we get to Lint, but not in this PR because we'd lose the majority of the comments and discussions. A new PR will follow today with that change.

docs/supported-tools.md
rock3r: Ok, a bit clearer now :) Well, if your concern is just that we should not mandate _how_ the Detekt plugin is added, we may just want to rephrase the text before the list from:

```
In order to use Detekt in your project, you need to:
```

to:

```
In order to use Detekt, you need to manually add it to **all** your Kotlin projects.
You can refer to the [official documentation](https://github.com/arturbosch/detekt/#gradlegroovy)
for further details. Note that you should _not_ add the `detekt` closure to your
`build.gradle`s, unlike what the official documentation says. The `detekt` closure
in the `staticAnalysis` configuration gets applied to all Kotlin modules automatically.

In most common cases, adding Detekt to a project boils down to three simple steps:
[...]
```

It is more verbose (mostly because of the not adding `detekt` stuff) than just having what we have now, but if you refer to the official docs you _must_ point out not to add the `detekt` configuration — lest things not work as expected. That was the reason why, as I explained earlier, I didn't refer users to the official docs (which by the way make a very poor job at explaining what to do).

docs/supported-tools.md
rock3r: I will postpone this to _after_ I have split out the tools docs, one per file, so I don't have to go back and change things again. I need #57 merged, and then this one merged, to be able to proceed with this.

docs/supported-tools.md
rock3r: Continues in https://github.com/novoda/gradle-static-analysis-plugin/pull/54#discussion_r164992250

docs/advanced-usage.md
rock3r: Yes. Good spot! I hadn't looked at the Detekt config yet :)

docs/advanced-usage.md
rock3r: Ditto 👍

docs/supported-tools.md
rock3r: Yes! I totally forgot about it!

docs/supported-tools.md
tasomaniac: More verbose but at the same time more maintainable. 
I personally like the last one you suggested in above comment.

docs/supported-tools.md
tasomaniac: That sounds good.

docs/supported-tools.md
rock3r: a38c70b748b653e46cae764d2025c26faa061f77

docs/advanced-usage.md
rock3r: 21430dcd8a758a6a56d1690c4363aaedd7f053ca

docs/advanced-usage.md
rock3r: 21430dcd8a758a6a56d1690c4363aaedd7f053ca

README.md
rock3r: cb7a7d4eae55d8d31d7642abf765a49a2c755257

README.md
rock3r: f9b8f07d553469d2540395e3b5a9739abc39ed25

rock3r:
I have added some explanation on how to exclude files from Detekt. Unfortunately the official docs do not really explain that, so we have to do it ourselves.

docs/supported-tools.md
rock3r: fadaac3f765705d44d1c12b14ec0d7f8d643688c

rock3r:
@tasomaniac @mr-archano @tobiasheine all comments addressed. Will start working on splitting out the docs for each tool in their own file next.

tobiasheine:
@rock3r Do you wanna do the split as part of this PR or can we merge it?

tasomaniac:
Let's just merge it. 👍

Issue #67, opened 2018-02-01
Re-visit definition/structure of internal packages


rock3r:
Since v0.5.0 we have exposed the `ViolationsEvaluator` interface in the public API (see #45). The interface has a single method, which takes in a `Set<Violations>`, but the `Violations` class is in the `internal` package. It should be moved up one level to be officially part of the public API.

We should also double check if anything is in the wrong package (@mr-archano mentioned all the integration tests being in the `internal` package, which is not necessary). We may also want to annotate non-public API with the [`@RestrictTo`](https://developer.android.com/studio/write/annotations.html#restrict) annotation to formally mark them as private (since Groovy has no `internal` modifier like Kotlin, it's the best we can do)

mr-archano:
I don't feel like we need to pull in an artifact from an _alien_ echosystem for an annotation. We can create our own `@Beta`, `@Incubating` or whatever. They are just markers anyway.

rock3r:
I'd agree, the only difference is that IDEs generally understand the Android Support annotations and mark things in the editor with warnings.

tasomaniac:
@rock3r are you sure that `@RestrictTo` works with gradle sources? It is used by Android Lint and works for source files as far as I know. Not for build scripts.

rock3r:
@tasomaniac I _think_ it should work but I am not sure (it's mostly about the IDE supporting it, and it usually picks up annotations such as nullability in build scripts). Hence why I'd suggest we try it, if it works it's fine, if it doesn't we don't lose anything, we just keep going as we're already doing.

tasomaniac:
That sounds good to me. Also really curious.

mr-archano:
We have addressed the most urgent issue via #69, we'll keep this issue open to keep an eye on the evolution of the internals.

tasomaniac:
Going ahead and closing this issue since the mention class was moved out from `internal` and the rest in there are really internal

PR #69, opened 2018-02-01
[PT-523] - Move Violations out of the internal package

tobiasheine:
[PT-523](https://novoda.atlassian.net/browse/PT-523)

### Description
With version 0.5 `Violations` are part of the public API, but they're still located in the `internal` package. Click [here](https://github.com/novoda/gradle-static-analysis-plugin/issues/67) for more details.

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/pmd/CollectPmdViolationsTask.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/pmd/CollectPmdViolationsTask.groovy
@@ -1,7 +1,7 @@
 package com.novoda.staticanalysis.internal.pmd
 
-import com.novoda.staticanalysis.internal.CollectViolationsTask
 import com.novoda.staticanalysis.Violations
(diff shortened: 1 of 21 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/internal/pmd/CollectPmdViolationsTask.groovy
```diff
@@ -1,7 +1,7 @@
 package com.novoda.staticanalysis.internal.pmd
 
 import com.novoda.staticanalysis.internal.CollectViolationsTask
-import com.novoda.staticanalysis.internal.Violations
+import com.novoda.staticanalysis.Violations
```
tasomaniac: Can you re-format all the files, I believe this import should come first.

tobiasheine: Such a eagle 👁

PR #73, opened 2018-02-02
[PT-508]: Prepare gradle plugin repo release

tobiasheine:
### Description
This PR prepares the release to to the [Gradle Plugins Repositiory](https://plugins.gradle.org/) using use [Gradle Publish Plugin](https://plugins.gradle.org/docs/publish-plugin). 

Next steps would be:
- approve the PR 
- publish via `publishPlugins` from my local machine to see whether it works
- update `.gradle/gradle.properties` on the CI with the credentials
- merge this PR

### Considerations
- Update readme to mention Kotlin Projects
- Add instructions how to apply the plugin from the Gradle Plugin Repository
- Configure the Gradle Publish Plugin


gradle/publish.gradle
```diff
@@ -1,7 +1,9 @@
 version = '0.5.1'
+group = 'com.novoda.gradle-static-analysis-plugin'
 String tag = "v$project.version"
 groovydoc.docTitle = 'Static Analysis Plugin'
 
+apply plugin: "com.gradle.plugin-publish"
```
mr-archano: single quotes pls

gradle/publish.gradle
```diff
@@ -1,7 +1,9 @@
 version = '0.5.1'
+group = 'com.novoda.gradle-static-analysis-plugin'
```
mr-archano: it seems like this is overlapping with the properties we use to publish to bintray (`groupId` + `artifactId`), what if we define those as local vars here and we compose `group` as convolution (ie: `group = "$groupId.$artifactId"`)?

tobiasheine: I added them as constants to be reused all over the script

README.md
```diff
@@ -47,12 +47,17 @@ buildscript {
         classpath 'com.novoda:gradle-static-analysis-plugin:0.5.1'
     }
 }
-```
 
-You can then apply the plugin via:
+apply plugin: 'com.novoda.static-analysis'
+```
+
```
rock3r: Maybe missing an OR, the two ways seem both required right now

README.md
```diff
@@ -47,12 +47,17 @@ buildscript {
...
-You can then apply the plugin via:
+apply plugin: 'com.novoda.static-analysis'
+```
+        
+Apply the plugin from the [Gradle Plugins Repository](https://plugins.gradle.org/):
 
 ```gradle
-apply plugin: 'com.novoda.static-analysis'
+plugins {
+    id 'com.novoda:gradle-static-analysis-plugin' version ''
```
rock3r: ⚠️ missing the `version` value!

gradle/publish.gradle
```diff
@@ -33,6 +41,20 @@ githubPages {
     }
 }
 
+pluginBundle {
```
rock3r: Maybe add a `displayName` too?

tobiasheine: Mmh there is a `displayName`

```
displayName = 'Gradle static analysis plugin'
```

rock3r: I am blind LOL

build.gradle
```diff
@@ -4,11 +4,15 @@ subprojects {
     buildscript {
         repositories {
             jcenter()
+            maven {
```
tasomaniac: nit: these 3 lines can be replaced by `gradlePluginPortal()`
If it is not available, we can update gradle version to `4.4.1` to replace it. 

Can be done later too.

README.md
```diff
@@ -47,12 +47,17 @@ buildscript {
...
-You can then apply the plugin via:
+apply plugin: 'com.novoda.static-analysis'
+```
+        
+or from the [Gradle Plugins Repository](https://plugins.gradle.org/):
 
 ```gradle
-apply plugin: 'com.novoda.static-analysis'
+plugins {
+    id 'com.novoda:gradle-static-analysis-plugin' version '0.5.1'
```
tasomaniac: :x: What about keeping both consistent? 
I think this should be `com.novoda.static-analysis`.
Also `gradle` and `plugin` words are redundant in this context. Having them in the repo name makes sense tho but maybe not in this name here.

gradle/publish.gradle
```diff
@@ -1,14 +1,22 @@
+ext {
+    GROUP_ID = 'com.novoda'
+    ARTIFACT_ID = 'gradle-static-analysis-plugin'
+    WEBSITE = 'https://github.com/novoda/gradle-static-analysis-plugin'
+}
+
 version = '0.5.1'
+group = "$GROUP_ID.$ARTIFACT_ID"
```
tasomaniac: I wonder if this `group` variable is used. 🤔

gradle/publish.gradle
```diff
@@ -33,6 +41,20 @@ githubPages {
...
 
+pluginBundle {
+    website = WEBSITE
+    vcsUrl = WEBSITE
+    description = 'A Gradle plugin to easily apply the same setup of static analysis tools across different Android, Java or Kotlin projects.'
+    tags = ['java', 'kotlin', 'android', 'static analysis', 'code quality']
+
+    plugins {
+        gradleStaticAnalysisPlugin {
+            id = "$GROUP_ID.$ARTIFACT_ID"
```
tasomaniac: Same comment here. 
Also in jCenter, we have the full name `gradle-static-analysis-plugin` as id because jCenter contains all libraries. So having gradle and plugin names make sense. 

But Gradle Plugin Portal only consist of those. So we should really have just `com.novoda.static-analysis`

README.md
mr-archano: @tasomaniac has a good point here: I believe this needs to follow the naming we use with `apply` method, so in this case `com.novoda.static-analysis`.

gradle/publish.gradle
```diff
@@ -1,14 +1,22 @@
+ext {
+    GROUP_ID = 'com.novoda'
```
mr-archano: these are all variables, not constants, so we can avoid to use the ALL_CAPS format here

gradle/publish.gradle
mr-archano: I'm not sure either: the [documentation](https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_markers) mentions that only `plugin.id` and `plugin.version` are used to index an entry in the portal, but then the mention of [`java-gradle-plugin`](https://guides.gradle.org/implementing-gradle-plugins/#plugin-development-plugin) makes everything more confused

gradle/publish.gradle
tobiasheine: that makes sense 👍

README.md
tobiasheine: yep, I agree 💯

gradle/publish.gradle
tobiasheine: True that, I followed this [documentation](https://plugins.gradle.org/docs/publish-plugin) which also just mentions the `version`. Will remove the `group` for now.

build.gradle
tobiasheine: done 👍

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -2,4 +2,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.3.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4.1-all.zip
```
mr-archano: have you used the `wrapper` task to do this? I'm just asking because sometimes that updates also the wrapper scripts and such

tobiasheine: No I didn't. Will do that 👍

tobiasheine: It changed the path from `distributionUrl=https\://services.gradle.org/distributions/gradle-4.4.1-all.zip` to `distributionUrl=https\://services.gradle.org/distributions/gradle-4.4.1-bin.zip` 🤔

mr-archano: we actually want `all.zip`

build.gradle
```diff
@@ -23,3 +25,8 @@ subprojects {
         }
     }
 }
+
+task wrapper(type: Wrapper) {
+    gradleVersion = '4.4.1'
```
mr-archano: In future we can just bump the version here and run `./gradlew wrapper` to update the Gradle distribution

tobiasheine:
Just released it manually:
https://plugins.gradle.org/plugin/com.novoda.static-analysis

mr-archano:
@tobiasheine it would be good to add a follow-up task to update the credentials on the CI before next release

PR #76, opened 2018-02-06
PT-548: Consume gradle plugins repo credentials from secrets.properties

tobiasheine:
### Description 
According to the [documentation](https://plugins.gradle.org/docs/submit) the credentials for the gradle plugins repository need to be put to the `gradle.properties`. But this would mean we'd depend on a specific machine.

I checked the [source code](https://plugins.gradle.org/m2/com/gradle/publish/plugin-publish-plugin/0.9.1/?_ga=2.243434636.886326088.1517819728-636804111.1514966621) of the plugin (the last version which is available as open source) and found out that we can overwrite this by setting a system property `com.gradle.login.properties.file` and link to another file instead.

Since we're using the `build-properties` plugin we decided to use the `secrets.properties`. The goal of this PR is to set `com.gradle.login.properties.file` to the path of `secrets.properties` if it is available and contains the needed credentials.

### Considerations
- If `secrets.properties` exists and it contains `gradle.publish.key`, `gradle.publish.secret` set `com.gradle.login.properties.file` to it's absolute path

### DIFF
--- a/gradle/publish.gradle
+++ b/gradle/publish.gradle
@@ -69,8 +69,6 @@ prepareGhPages.dependsOn groovydoc
 publishGhPages.dependsOn prepareGhCredentials
 
 task prepareGradlePluginsRepoRelease {
-    description = "Prepare the release to the Gradle Plugins Repository"
-    group = 'release'
     doLast {
         // force evaluation of properties to check whether they're there or not
         buildProperties.secrets['gradle.publish.key'].string
--- a/gradle/publish.gradle
+++ b/gradle/publish.gradle
@@ -84,11 +84,11 @@ private static void validateSecrets(File secretsProperties) {
     def props = new Properties()
     secretsProperties.withInputStream { props.load(it) }
 
-    if (!props.hasProperty('gradle.publish.key')) {
+    if (!props.containsKey('gradle.publish.key')) {
         throw new GradleException('Could not find find gradle.publish.key in secrets.properties')
     }
 
-    if (!props.hasProperty('gradle.publish.secret')) {
+    if (!props.containsKey('gradle.publish.secret')) {
         throw new GradleException('Could not find find gradle.publish.secret in secrets.properties')
     }
 }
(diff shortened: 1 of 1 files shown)


gradle/publish.gradle
```diff
@@ -66,10 +68,35 @@ task prepareGhCredentials {
 prepareGhPages.dependsOn groovydoc
 publishGhPages.dependsOn prepareGhCredentials
 
+task prepareGradlePluginsRepoRelease {
+    description = "Prepare the release to the Gradle Plugins Repository"
+    group = 'release'
+    doLast {
+        def secretsProperties = rootProject.file('secrets.properties')
+        if (secretsProperties.exists()) {
```
mr-archano: I believe we can just assume that we *always* expect the credentials to be part of `secrets.properties` and let the `gradle-build-properties-plugin` to deal with that. This task's action would become:
```
// force evaluation of properties to check whether they're there or not
buildProperties.secrets['gradle.publish.key'].string
buildProperties.secrets['gradle.publish.secret'].string
System.properties['com.gradle.login.properties.file'] = rootProject.file('secrets.properties')
```

gradle/publish.gradle
```diff
@@ -66,10 +68,35 @@ task prepareGhCredentials {
...
+            validateSecrets(secretsProperties)
```
tasomaniac: validateSecrets already checks essentially that the file exists. We can simplify by removing the if statement.

gradle/publish.gradle
```diff
@@ -66,10 +68,35 @@ task prepareGhCredentials {
 prepareGhPages.dependsOn groovydoc
 publishGhPages.dependsOn prepareGhCredentials
 
+task prepareGradlePluginsRepoRelease {
+    description = "Prepare the release to the Gradle Plugins Repository"
+    group = 'release'
```
tasomaniac: `nit`: When `group` is set, it shows up in the IDE. It is more for tasks that are meant to be run manually. Since this is a helper, `group` is not necessary. Also `description`

PR #79, opened 2018-02-09
Android lint variant filter support

### LINKED ISSUE #66: includeVariants support for Android Lint setup
In the newly added Android Lint feature, we just run `lint` which runs lint against the whole project with all variants. This also needs a compilation of all variants because lint runs on the compiled code, not on the source. 

For this reason, lint can take really too much time. 

Suggestion: Add support for `includeVariants` to just run on certain variants.

And also we can consider supporting `includeVariants` on upper level to be used on all tools.

### LINKED ISSUE #77: Timing issue with lint configuration
Our Android lint configurator just early return when the Android plugin is not found yet. 

I was trying to modify out integration in a project to add lint but I realized that it silently fails and just does configure lint. 

The reason is that the `static-analysis` plugin is applied in a `subprojects` closure in the root `build.gradle` when it is applied, the android plugin is not configured yet. 

Just like other configurators, using `plugins.withId` may fix this because closure provided with `withId` will be called once the android plugin is applied.

tasomaniac:
Fixes #66 
Fixes #77 

## Problem 

The lint setup didn't have filters around Android variants. In a big project with 10+ variants, the `evaluateViolations` task take too long. (around 20 mins)

## Solution

This PR make `lintOptions` variant aware. 

- Introduced a `trait` to encapsulate Android variant awareness in Configurators.
  - A follow-up will be done to this to make it cleaner.
- Instead of skipping non-android projects, use `project.plugins.withId('com.android.application')` to wait for potential plugin addition. This will fix the timing issue described in #77 
- When lint runs on a single variant, the output names change. That's why introduced `Nullable` variant param to task creation and change the report file path accordingly. 

**Note:** When filter is not provided, it will keep the old behavior. When filter is used, we may get duplicated results when the same problem is available in an intersection of the sourceSets of multiple variants. This problem is available in all tasks and fix will be introduced later. 

## Tests

Lots of tests added including the duplicate number behavior.
A test is also added to check if we really ignore problems when the variant is ignored.

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
@@ -5,32 +5,26 @@ import org.gradle.api.NamedDomainObjectSet
 
 trait VariantAware {
 
-    private boolean hasFilter = false
-    Closure<Boolean> includeVariantsFilter = { true }
+    Closure<Boolean> includeVariantsFilter
 
-    void setIncludeVariantsFilter(Closure<Boolean> includeVariantsFilter) {
-        this.hasFilter = true
-        this.includeVariantsFilter = includeVariantsFilter
-    }
-
-    boolean getHasFilter() {
-        hasFilter
+    final variantSelector = { DomainObjectSet variants ->
+        includeVariantsFilter != null ? variants.matching { includeVariantsFilter(it) } : variants
     }
 
     DomainObjectSet<Object> getFilteredApplicationVariants() {
-        project.android.applicationVariants.matching { includeVariantsFilter(it) }
+        variantSelector(project.android.applicationVariants)
     }
 
     NamedDomainObjectSet<Object> getFilteredApplicationAndTestVariants() {
-        getAllVariants(project.android.applicationVariants).matching { includeVariantsFilter(it) }
+        variantSelector(getAllVariants(project.android.applicationVariants))
     }
 
     DomainObjectSet<Object> getFilteredLibraryVariants() {
-        project.android.libraryVariants.matching { includeVariantsFilter(it) }
+        variantSelector(project.android.libraryVariants)
     }
 
     NamedDomainObjectSet<Object> getFilteredLibraryAndTestVariants() {
-        getAllVariants(project.android.libraryVariants).matching { includeVariantsFilter(it) }
+        variantSelector(getAllVariants(project.android.libraryVariants))
     }
 
     private NamedDomainObjectSet<Object> getAllVariants(variants1) {
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
@@ -1,0 +1,24 @@
+package com.novoda.staticanalysis.internal
+
+import org.gradle.api.NamedDomainObjectSet
+
+trait VariantAware {
+
+    Closure<Boolean> includeVariantsFilter = { true }
+
+    NamedDomainObjectSet<Object> getFilteredApplicationVariants() {
+        getAllVariants(project.android.applicationVariants).matching { includeVariantsFilter(it) }
+    }
+
+    NamedDomainObjectSet<Object> getFilteredLibraryVariants() {
+        getAllVariants(project.android.libraryVariants).matching { includeVariantsFilter(it) }
+    }
+
+    private NamedDomainObjectSet<Object> getAllVariants(variants1) {
+        NamedDomainObjectSet<Object> variants = project.container(Object)
+        variants.addAll(variants1)
+        variants.addAll(project.android.testVariants)
+        variants.addAll(project.android.unitTestVariants)
+        return variants
+    }
+}
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
@@ -1,16 +1,25 @@
 package com.novoda.staticanalysis.internal
 
+import org.gradle.api.DomainObjectSet
 import org.gradle.api.NamedDomainObjectSet
 
 trait VariantAware {
 
     Closure<Boolean> includeVariantsFilter = { true }
 
-    NamedDomainObjectSet<Object> getFilteredApplicationVariants() {
+    DomainObjectSet<Object> getFilteredApplicationVariants() {
+        project.android.applicationVariants.matching { includeVariantsFilter(it) }
+    }
+
+    DomainObjectSet<Object> getFilteredApplicationAndTestVariants() {
         getAllVariants(project.android.applicationVariants).matching { includeVariantsFilter(it) }
     }
 
-    NamedDomainObjectSet<Object> getFilteredLibraryVariants() {
+    DomainObjectSet<Object> getFilteredLibraryVariants() {
+        project.android.libraryVariants.matching { includeVariantsFilter(it) }
+    }
+
+    DomainObjectSet<Object> getFilteredLibraryAndTestVariants() {
         getAllVariants(project.android.libraryVariants).matching { includeVariantsFilter(it) }
     }
 
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
@@ -7,7 +7,7 @@ trait VariantAware {
 
     Closure<Boolean> includeVariantsFilter
 
-    final variantSelector = { DomainObjectSet variants ->
+    final variantSelector = { variants ->
         includeVariantsFilter != null ? variants.matching { includeVariantsFilter(it) } : variants
     }
 
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
@@ -5,13 +5,23 @@ import org.gradle.api.NamedDomainObjectSet
 
 trait VariantAware {
 
+    private boolean hasFilter = false
     Closure<Boolean> includeVariantsFilter = { true }
+
+    void setIncludeVariantsFilter(Closure<Boolean> includeVariantsFilter) {
+        this.hasFilter = true
+        this.includeVariantsFilter = includeVariantsFilter
+    }
+
+    boolean getHasFilter() {
+        hasFilter
+    }
 
     DomainObjectSet<Object> getFilteredApplicationVariants() {
         project.android.applicationVariants.matching { includeVariantsFilter(it) }
     }
 
-    DomainObjectSet<Object> getFilteredApplicationAndTestVariants() {
+    NamedDomainObjectSet<Object> getFilteredApplicationAndTestVariants() {
         getAllVariants(project.android.applicationVariants).matching { includeVariantsFilter(it) }
     }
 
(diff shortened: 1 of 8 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
```diff
@@ -0,0 +1,43 @@
+package com.novoda.staticanalysis.internal
+
+import org.gradle.api.DomainObjectSet
+import org.gradle.api.NamedDomainObjectSet
+
+trait VariantAware {
+
+    private boolean hasFilter = false
```
mr-archano: why don't we keep out of the `trait` API and we just keep the `getFiltered*()`? this way we can choose in the configurator how to deal with that, for instance only Lint needs to know whether the filter is set or not, right?

plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
```diff
@@ -0,0 +1,43 @@
...
+    Closure<Boolean> includeVariantsFilter = { true }
```
mr-archano: this can be just `Closure<Boolean> getIncludeVariantsFilter()` that we'll leave to the configurators to implement.

tasomaniac: `setter` is also used in subclasses. I can of course do this but the beauty of a trait is to encapsulate the variant behavior all here.
If I implement `getIncludeVariantsFilter` in the subclasses, then I would need a property there.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
tasomaniac: `trait` allows me to have implementation and fields here. I don't want to lose that. 

The other option is to make it nullable. Let me try and see how that goes actually. `hasFilter` is not necessary.

tasomaniac: Oh, addressing here, I pushed to `clean-up` branch. That looks much nicer.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
```diff
@@ -28,16 +30,30 @@ class LintConfigurator implements Configurator {
...
+            project.plugins.withId('com.android.library') {
+                configureWithVariants(config, filteredLibraryVariants)
+            }
+
+        }
+    }
+
+    private void configureWithVariants(Closure config, DomainObjectSet variants) {
+        configureLint(config)
+        if (hasFilter) {
```
mr-archano: where `hasFilter` is coming from now? not sure how this is working now - if it's working even

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
```diff
@@ -46,32 +62,29 @@ class LintConfigurator implements Configurator {
...
+            def reportSuffix = taskSuffix ? "-$taskSuffix" : ''
+            task.xmlReportFile = xmlOutputFileFor(reportSuffix)
+            task.htmlReportFile = new File(defaultOutputFolder, "lint-results${reportSuffix}.html")
+            task.violations = violations
         }
     }
 
-    private File getXmlOutputFile() {
-        project.android.lintOptions.xmlOutput ?: new File(defaultOutputFolder, 'lint-results.xml')
+    private File xmlOutputFileFor(reportSuffix) {
```
mr-archano: I would prefer to push the report name directly and take the decision up two levels from here, so to keep it close to where we call `configureCollectViolationsTask()`

tasomaniac: I'm not sure what you mean by 2 level. Do mean the `with` clause above just after we create the task. The `xmlReportFile` property of the task is better to be configured while it is created.

tasomaniac:
retest this please

plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
```diff
@@ -0,0 +1,37 @@
+package com.novoda.staticanalysis.internal
+
+import org.gradle.api.DomainObjectSet
+import org.gradle.api.NamedDomainObjectSet
+
+trait VariantAware {
```
tobiasheine: Personally I would haven chosen composition over inheritance here since we mostly expose fields and methods to be shared.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
```diff
@@ -28,16 +30,30 @@ class LintConfigurator implements Configurator {
...
+        project.extensions.findByType(StaticAnalysisExtension).ext.lintOptions = { Closure config ->
+            project.plugins.withId('com.android.application') {
+                configureWithVariants(config, filteredApplicationVariants)
             }
-            configureLint(config)
-            configureToolTask()
+            project.plugins.withId('com.android.library') {
+                configureWithVariants(config, filteredLibraryVariants)
+            }
+
```
tobiasheine: nit: empty line

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
```diff
@@ -28,16 +30,30 @@ class LintConfigurator implements Configurator {
...
+        configureLint(config)
+        if (includeVariantsFilter != null) {
+            variants.all { configureCollectViolationsTask(it) }
+        } else {
+            configureCollectViolationsTask()
         }
     }
 
     private void configureLint(Closure config) {
+        project.android.lintOptions.ext.includeVariants = { Closure<Boolean> filter ->
```
tobiasheine: @mr-archano 
Could u explain us what exactly happens in this closure?

mr-archano: I'm not sure what you're asking @tobiasheine (as I didn't write this), but I will try to explain what I can see going on here: L54 adds an extension method to `lintOptions` that takes a `filter` predicate, used by `VariantAware.variantSelector` to actually filter in only the variants for which `filter` is evaluated as `true`

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
mr-archano: what I mean is that instead of adding this suffix logic in two places you can move the decision of deciding what task name and what report name to use up to L36 and L38.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
```diff
@@ -0,0 +1,37 @@
...
+
+    Closure<Boolean> includeVariantsFilter
+
+    final variantSelector = { variants ->
```
mr-archano: can this just be a normal private method then? it's weird to see it defined  it like this...

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
```diff
@@ -46,32 +62,29 @@ class LintConfigurator implements Configurator {
...
+            task.xmlReportFile = xmlOutputFileFor(reportSuffix)
+            task.htmlReportFile = new File(defaultOutputFolder, "lint-results${reportSuffix}.html")
+            task.violations = violations
         }
     }
 
-    private File getXmlOutputFile() {
-        project.android.lintOptions.xmlOutput ?: new File(defaultOutputFolder, 'lint-results.xml')
+    private File xmlOutputFileFor(reportSuffix) {
+        project.android.lintOptions.xmlOutput ?: new File(defaultOutputFolder, "lint-results${reportSuffix}.xml")
```
mr-archano: does this mean that when the user sets `xmlOutput` then every variant task will output to the same report file?

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
tasomaniac: Right, he was asking because this is a copy paste from the other ones.
I was also wondering how this works exactly. One question comes to mind, there is no mention of the input of the closure. We expect the input to be a variant but I guess there is no limit? Can it be anything?

plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantAware.groovy
tasomaniac: Method didn't work how I wanted it to be with this trait thingy. I guess I can do Tobi's suggestion and make this a real collaborator class.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
tasomaniac: No idea. ☺ Good question. Only way to find out is to try. I don't think the documentation mentions that.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
tobiasheine: I understand that we define a extension `includeVariants` with a `Closure<Boolean>` and what it is used for.
What is confusing to me is the assignment `includeVariantsFilter = filter` within the closure or when it is executed.  Is it like a callback that is triggered when the user specifies the extension?

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
tasomaniac: It overrides the files. If you have xmlOuput defined, you get a single output no matter which lint task you run. So this seems to be the right way to do. But also we should do it for HTML report actually. Same happens for that too.

tasomaniac:
retest this please

tasomaniac:
@mr-archano please have another look.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/VariantFilter.groovy
```diff
@@ -0,0 +1,43 @@
+package com.novoda.staticanalysis.internal
+
+import org.gradle.api.DomainObjectSet
+import org.gradle.api.NamedDomainObjectSet
+import org.gradle.api.Project
+
+final class VariantFilter {
```
mr-archano: ✨

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
```diff
@@ -46,32 +56,41 @@ class LintConfigurator implements Configurator {
         }
     }
 
-    private void configureToolTask() {
-        // evaluate violations after lint
-        def collectViolations = createCollectViolationsTask(violations)
+    private void configureWithVariants(DomainObjectSet variants) {
+        if (variantFilter.includeVariantsFilter != null) {
+            variants.all {
+                configureCollectViolationsTask(it.name, "lint-results-${it.name}")
```
mr-archano: me happy! 🐶

plugin/src/main/groovy/com/novoda/staticanalysis/internal/lint/LintConfigurator.groovy
mr-archano: that's weird behaviour. I agree that this is in line with the default reporting, but I wonder if we should provide a better way to handle that when using variant-aware lint tasks (ie: log a warning/error, or ignore that, or try to generate variant-aware reports from `xmlOutput`).

tasomaniac: I agree. We can talk to Android team about this. 
I don't think it is possible to do variant-aware custom `xmlOutput` with the current lintOptions in Android build tools. It is because this configuration is done only once in the configuration phase. Whatever you do overrides the file name. You can only have 1 single name for the output. If it is null, they generate file names with variant name in it.

PR #80, opened 2018-02-09
Clean up variant aware code

tasomaniac:
⚠️ Depends in #79 

`DomainObjectSet.all` loop is live and thanks to that usually `afterEvaluate` is not necessary because the closure will be called later (just after) the variants have been added. 

`getAllVariants` was merging android variants with test and unitTest variants. This was breaking the liveliness of the set. That's why we needed `afterEvaluate`.

Since the variant aware code is extracted away, this PR cleans this up by calling `all` on variants one by one instead of merging them. 

And thanks to that, the for looping over variants happen in the base class. This means that the implementations for Checkstyle, Findbugs and PMD are cleaner because they configure 1 variant at a time.

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/checkstyle/CheckstyleConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/checkstyle/CheckstyleConfigurator.groovy
@@ -48,23 +48,21 @@ class CheckstyleConfigurator extends CodeQualityConfigurator<Checkstyle, Checkst
     }
 
     @Override
-    protected void configureAndroidProject(NamedDomainObjectSet variants) {
+    protected void configureAndroidVariant(variant) {
         project.with {
-            variants.all { variant ->
-                variant.sourceSets.each { sourceSet ->
-                    def taskName = "checkstyle${sourceSet.name.capitalize()}"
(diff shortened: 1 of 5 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/internal/checkstyle/CheckstyleConfigurator.groovy
```diff
@@ -48,23 +48,21 @@ class CheckstyleConfigurator extends CodeQualityConfigurator<Checkstyle, Checkst
...
-                    def taskName = "checkstyle${sourceSet.name.capitalize()}"
-                    Checkstyle task = tasks.findByName(taskName)
-                    if (task == null) {
-                        task = tasks.create(taskName, Checkstyle)
-                        task.with {
-                            description = "Run Checkstyle analysis for ${sourceSet.name} classes"
-                            source = sourceSet.java.srcDirs
-                            classpath = files("$buildDir/intermediates/classes/")
-                        }
+            variant.sourceSets.each { sourceSet ->
```
tasomaniac: Just 1 level indentation deleted, no change.

tasomaniac:
retest this please

tasomaniac:
retest this please

PR #82, opened 2018-02-13
Fix properties in release script

tasomaniac:
The release script failed with `MissingPropertyException` while trying to set the property. 

It seems that `setProperty` is meant to be used when the property is already available. If not, it throws an exception. The following map interface through `properties` map seems to be the way to do it. 

While doing so, I also used or operator instead of the or method. Since it has less parenthesis, I think it is easier to see what's going on. I can revert that if wanted.

### DIFF
--- a/gradle/publish.gradle
+++ b/gradle/publish.gradle
@@ -72,8 +72,8 @@ task prepareGradlePluginsRepoRelease {
     doLast {
         String key = buildProperties.env['GRADLE_PLUGINS_REPO_KEY'].or(buildProperties.secrets['gradle.publish.key']).string
         String secret = buildProperties.env['GRADLE_PLUGINS_REPO_SECRET'].or(buildProperties.secrets['gradle.publish.secret']).string
-        project.setProperty('gradle.publish.key', key)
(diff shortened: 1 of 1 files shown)


gradle/publish.gradle
```diff
@@ -70,10 +70,10 @@ publishGhPages.dependsOn prepareGhCredentials
 
 task prepareGradlePluginsRepoRelease {
     doLast {
-        String key = buildProperties.env['GRADLE_PLUGINS_REPO_KEY'].or(buildProperties.secrets['gradle.publish.key']).string
-        String secret = buildProperties.env['GRADLE_PLUGINS_REPO_SECRET'].or(buildProperties.secrets['gradle.publish.secret']).string
-        project.setProperty('gradle.publish.key', key)
```
mr-archano: I assumed this was working as @tobiasheine did the last release to Gradle Plugins Portal; does it mean that it wasn't the case?

tasomaniac: In that release properties were read from a file. This change was not tested.

tobiasheine: @mr-archano the last release was not yet automated. This is the first one.

Issue #75, opened 2018-02-05
How do I disable failOnErrors?


tony19:
I'd like to set `failOnErrors false` so that the Gradle task doesn't fail when bugs are detected. How can I do that?

I tried:
```
staticAnalysis {
 failOnErrors false
  //...
}
```

and
```
staticAnalysis {
  penalty {
    failOnErrors false
    maxErrors 0
    maxWarnings 0
  }
  //...
}
```
and
```
staticAnalysis {
  penalty {
    none
  }
  //...
}
```

tasomaniac:
Thanks for the feedback. We might consider adding such option to dsl to be used in the default evaluator. 

To be able to do this right now, you would need a custom evaluator. More information here: https://github.com/novoda/gradle-static-analysis-plugin/blob/master/docs/incubating/custom-evaluator.md#custom-violations-evaluator-incubating

mr-archano:
@tony19 the last thing you posted is almost correct, in fact the extension provides 3 built-in (undocumented :( ) profiles for `penalty` (see [`StaticAnalysisExtension.groovy#L11-L24`](https://github.com/novoda/gradle-static-analysis-plugin/blob/master/plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisExtension.groovy#L11-L24)). In your case it's enough to write:
```
staticAnalysis {
    penalty none
}
```

mr-archano:
The alternative is to write a custom evaluator as @tasomaniac mentioned, that is a function that receives a `Set<Violations>` and act upon them.

tasomaniac:
I forgot `penalty`, that seems more appropriate.

tony19:
Thanks! `penalty none` works for me.

PR #85, opened 2018-02-14
More documentation around penalty

tasomaniac:
Issue #75 showed us we need more documentation around available convenience methods for penalty object.

Here is some.

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -31,6 +31,21 @@ fail or not.
 If you don't specify a `penalty` configuration, the plugin will use the [default threshold values][penaltyextensioncode], which are to
 allow any warning, but break the build on any error.
 
+It is also possible to use below convenience configuration to control failure thresholds
+
+```gradle
+staticAnalysis {
+    // no failure
+    penalty none
+    
+    // fail only on errors (default)
+    penalty failOnErrors
+    
+    // fail fast
+    penalty failOnWarnings
+}
+```
+
 ## Improve the report with a base URL
 Build logs will show an overall report of how many violations have been found during the analysis and the links to
 the relevant HTML reports, for instance:
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -7,7 +7,7 @@ the build will not fail, which is very useful for legacy projects.
  * [Configurable failure thresholds](#configurable-failure-thresholds)
  * [Improve the report with a base URL](#improve-the-report-with-a-base-URL)
  * [Add exclusions with `exclude` filters](#add-exclusions-with-exclude-filters)
- * [Add exclusions with Android build variants](#add-exclusions-with-Android-build-variants)
+ * [Add exclusions with Android build variants](#add-exclusions-with-android-build-variants)
  * [Custom violations evaluator (**incubating**)](incubating/custom-evaluator.md#custom-violations-evaluator-incubating)
 
 ---
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -31,20 +31,35 @@ fail or not.
 If you don't specify a `penalty` configuration, the plugin will use the [default threshold values][penaltyextensioncode], which are to
 allow any warning, but break the build on any error.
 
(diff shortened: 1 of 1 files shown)


docs/advanced-usage.md
```diff
@@ -31,6 +31,21 @@ fail or not.
...
 
+It is also possible to use below convenience configuration to control failure thresholds
+
+```gradle
+staticAnalysis {
+    // no failure
+    penalty none
+    
+    // fail only on errors (default)
+    penalty failOnErrors
```
mr-archano: Maybe it's nitpicking, but I'd rather split this a little bit, as people might be confused by the example. What about:
>
> It's also possible to use one of the default `penalty` profiles instead of specifying thresholds explicitly as follows:
<table>
<tr><td><pre>staticAnalysis {</br>&nbsp;&nbsp;penalty none</br>}</pre></td><td>In this case the build won't fail no matter how many violations (warnings or error) are found</><tr>
<tr><td><pre>staticAnalysis {</br>&nbsp;&nbsp;penalty failOnErrors</br>}</pre></td><td>Use this if you want to fail fast: any warning or this is the default policy and will break the build if any error is found. Warnings instead are logged but ignored during the evaluation.</><tr>
<tr><td><pre>staticAnalysis {</br>&nbsp;&nbsp;penalty failOnWarnings</br>}</pre></td><td>This is the default policy and will break the build if any warning or error is found.</><tr>
</table>

tasomaniac: I think it deserved a sub-section now. Updated. How does it look?

docs/advanced-usage.md
```diff
@@ -31,6 +31,36 @@ fail or not.
 If you don't specify a `penalty` configuration, the plugin will use the [default threshold values][penaltyextensioncode], which are to
 allow any warning, but break the build on any error.
 
+#### Default `penalty` profiles
```
rock3r: Two notes here:
1. Why is this an h4 and not h3? It is a sub-section of an h2, should be an h3 then
2. Can we put the various built-ins as a bulleted list? It's a bit unclear which text refers to which snippet right now

docs/advanced-usage.md
```diff
@@ -31,6 +31,36 @@ fail or not.
...
+Default policy. This will break the build if any error is found. Warnings instead are only logged and will not break the build.
+
+
+```gradle
+staticAnalysis {
+    penalty failOnWarnings
+}
+```
+
+Use this if you want to fail fast: the build will break if any warning or error is found.
```
rock3r: Given this is what it does then maybe we should rename `failOnWarnings` to `failFast`, it would seem much more clear as `failOnWarnings` doesn't clearly imply `failOnErrors` as a name IMO

mr-archano: good point @rock3r, we can address in this PR I guess

tasomaniac: Good point, I also saw that in the legacy version of this tool, there was an enum with values `NONE`, `ERRORS` and `ALL`. `ALL` was originally `WARNINGS`. 

So doing this will align that.

docs/advanced-usage.md
tasomaniac: First is done. I'm not sure how I will put `bullet`s here. Can you maybe modify to help this? You can even do that through the Github UI in this PR. Thanks.

rock3r: Sure! On it ;)

rock3r: 05c9492 done!

docs/advanced-usage.md
rock3r: Although GH thinks that 05c9492 addressed this, it didn't — please don't lose track of this rename 💃

tasomaniac: We have this now. https://github.com/novoda/gradle-static-analysis-plugin/issues/87

I was thinking to just change this in this PR but that requires a release. We should do it seprately.

docs/advanced-usage.md
tasomaniac: Thanks 👍 Much clear now. Especially the `(Default)` indicator.

PR #90, opened 2018-02-15
Verify detekt config

### LINKED ISSUE #70: Detekt: Exception when output is null
It seems that the output is not mandatory in detekt. When you don't have any output defined, it simply skips writing to files and only does console output. 

But when it is integrated through our plugin, we just throw `IllegalArgumentException` with a cryptic message. Instead, maybe we should validate it and require `output` to be define. 

<details>
Caused by: java.lang.IllegalArgumentException: path may not be null or empty string. path='null'
        at org.gradle.api.internal.file.AbstractBaseDirFileResolver.doResolve(AbstractBaseDirFileResolver.java:65)
        at org.gradle.api.internal.file.AbstractFileResolver.resolve(AbstractFileResolver.java:85)
        at org.gradle.api.internal.file.AbstractFileResolver.resolve(AbstractFileResolver.java:67)
        at org.gradle.api.internal.file.DefaultFileOperations.file(DefaultFileOperations.java:89)
        at org.gradle.api.internal.project.DefaultProject.file(DefaultProject.java:840)
        at org.gradle.api.Project$file$3.call(Unknown Source)
        at com.novoda.staticanalysis.internal.detekt.DetektConfigurator.createCollectViolationsTask(DetektConfigurator.groovy:67)
        at com.novoda.staticanalysis.internal.detekt.DetektConfigurator.configureToolTask(DetektConfigurator.groovy:61)
        at com.novoda.staticanalysis.internal.detekt.DetektConfigurator$_execute_closure1.doCall(DetektConfigurator.groovy:50)
</details>

tasomaniac:
Fixes #70 

# Problem

We don't verify the tools configurations but maybe we should in some cases. In this particular case, `output` is optional in detekt and in such case, it just prints to console and skips generating file reports. 

When this happens, our plugin fails with an exception while running with not so clear error message. 

# Solution

Check if the `output` is defined. 

# Tests

- Added a test case around the issue. 
- Moved configuration related tests together. 
- Little more boy-scouting in the test.

### DIFF
--- a/plugin/src/test/groovy/com/novoda/staticanalysis/internal/detekt/DetektIntegrationTest.groovy
+++ b/plugin/src/test/groovy/com/novoda/staticanalysis/internal/detekt/DetektIntegrationTest.groovy
@@ -29,7 +29,7 @@ class DetektIntegrationTest {
     }
 
     @Test
-    void givenOutputNotDefinedShouldFailBuildOnConfiguration() {
+    void shouldFailBuildOnConfigurationWhenNoOutputNotDefined() {
         def emptyConfiguration = detektWith("")
 
         def result = createProjectWithZeroThreshold(Fixtures.Detekt.SOURCES_WITH_WARNINGS)
--- a/plugin/src/test/groovy/com/novoda/staticanalysis/internal/detekt/DetektIntegrationTest.groovy
+++ b/plugin/src/test/groovy/com/novoda/staticanalysis/internal/detekt/DetektIntegrationTest.groovy
@@ -13,6 +13,9 @@ import static com.novoda.test.LogsSubject.assertThat
 @RunWith(Parameterized.class)
 class DetektIntegrationTest {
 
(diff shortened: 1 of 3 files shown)


plugin/src/test/groovy/com/novoda/staticanalysis/internal/detekt/DetektIntegrationTest.groovy
```diff
@@ -25,6 +28,26 @@ class DetektIntegrationTest {
         this.projectRule = projectRule
     }
 
+    @Test
+    void givenOutputNotDefinedShouldFailBuildOnConfiguration() {
```
mr-archano: the name of this test could be `shouldFailBuildOnConfigurationWhenOutputNotDefined`

tasomaniac: haha 😃The test name was like that in the beginning actually.

PR #93, opened 2018-02-16
Multiple config support for tools

tasomaniac:
Just like Android Lint, main tools (pmd, findbugs, checkstyle) also had the same problem where if the user configures them multiple times, the configuration fails. 

Tests were written first and then when `task already exists` problems happen, task creations are converted to `maybeCreate`

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/checkstyle/CheckstyleConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/checkstyle/CheckstyleConfigurator.groovy
@@ -80,9 +80,9 @@ class CheckstyleConfigurator extends CodeQualityConfigurator<Checkstyle, Checkst
     }
 
     private CollectCheckstyleViolationsTask createCollectViolationsTask(Checkstyle checkstyle, Violations violations) {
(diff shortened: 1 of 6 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/internal/checkstyle/CheckstyleConfigurator.groovy
```diff
@@ -80,9 +80,9 @@ class CheckstyleConfigurator extends CodeQualityConfigurator<Checkstyle, Checkst
     }
 
     private CollectCheckstyleViolationsTask createCollectViolationsTask(Checkstyle checkstyle, Violations violations) {
-        project.tasks.create("collect${checkstyle.name.capitalize()}Violations", CollectCheckstyleViolationsTask) { collectViolations ->
```
tasomaniac: Old version looks much better but unfortunately the same thing is not available for `maybeCreate`
And `with` does not make it better since `with` returns the last statement, I still need to return task manually.

mr-archano:
I don't understand what went wrong with this build - the error is about the missing resolution of the Detekt artifact, but I can't see any change in this PR affecting that.

@hal90002 could you retest this please?

PR #97, opened 2018-02-26
Document rules artefact usage

tobiasheine:
### Description
This PR adds the missing documentation of how to publish a set of rules as maven artefact and how to consume it from the plugin.

Also it references the novoda ruleset which has been released to bintray.

### Considerations
Just documentation, the feature has been already implemented earlier.

### DIFF
--- a/docs/advanced-usage.md
+++ b/docs/advanced-usage.md
@@ -8,6 +8,7 @@ the build will not fail, which is very useful for legacy projects.
  * [Improve the report with a base URL](#improve-the-report-with-a-base-URL)
  * [Add exclusions with `exclude` filters](#add-exclusions-with-exclude-filters)
  * [Add exclusions with Android build variants](#add-exclusions-with-android-build-variants)
+ * [Consume rules from artefact](#consume-rules-from-artefact)
  * [Custom violations evaluator (**incubating**)](incubating/custom-evaluator.md#custom-violations-evaluator-incubating)
 
 ---
@@ -131,3 +132,57 @@ staticAnalysis {
 Please note that this is not yet supported for Detekt.
 
 [penaltyextensioncode]: https://github.com/novoda/gradle-static-analysis-plugin/blob/master/plugin/src/main/groovy/com/novoda/staticanalysis/PenaltyExtension.groovy
+
+
+### Consume rules from artefact     
+
+In order to reuse your rules among multiple projects or to easily use an open source rule set, we added support for consuming the 
+rules for all supported tools from a maven artefact. An example of how to create and publish your rules as artefact can be 
+found [here](https://github.com/novoda/novoda/blob/master/scaffolding/build.gradle). Novoda open sourced their rules to [bintray](https://bintray.com/novoda/maven/static-analysis-rules) inlcluding
+rules for Checkstyle, PMD, FindBugs, detekt and Android Lint.
+
+
+To consume the rules from an artefact you need to configure the `rules` closure pointing to a maven artefact and use it when configuring the tools, e.g.
+ 
+```gradle
+staticAnalysis {
+    rules {
+        novoda {
+            maven 'com.novoda:static-analysis-rules:0.2'
+        }
+    }
+    
+    checkstyle {
+        toolVersion '8.8'
+        config rules.novoda['checkstyle-modules.xml']
+    }
+}
+```
+
+These are the necessary configurations for all supported tools:
+
+```gradle
+checkstyle {
+    toolVersion '8.8'
+    config rules.novoda['checkstyle-modules.xml']
+}
+
+pmd {
+    toolVersion '6.0.1'
+    ruleSetFiles = project.files(rules.novoda['pmd-rules.xml'].asFile().path)
+}
+
+findbugs {
+    excludeFilter rules.novoda['findbugs-excludes.xml'].asFile()
+}
+
+detekt {
+    profile('main') {
+        config = rules.novoda['detekt.yml'].asFile().path
+    }
+}
+
+lintOptions {
+    lintConfig = rules.novoda['lint-config.xml'].asFile()
+}
+```


docs/advanced-usage.md
```diff
@@ -8,6 +8,7 @@ the build will not fail, which is very useful for legacy projects.
  * [Improve the report with a base URL](#improve-the-report-with-a-base-URL)
  * [Add exclusions with `exclude` filters](#add-exclusions-with-exclude-filters)
  * [Add exclusions with Android build variants](#add-exclusions-with-android-build-variants)
+ * [Consume rules from artefact](#consume-rules-from-artefact)
```
mr-archano: isn't [`artifact` though](https://stackoverflow.com/questions/2487485/what-is-a-maven-artifact)?

docs/advanced-usage.md
```diff
@@ -131,3 +132,57 @@ staticAnalysis {
...
+    }
+    
+    checkstyle {
+        toolVersion '8.8'
+        config rules.novoda['checkstyle-modules.xml']
+    }
+}
+```
+
+These are the necessary configurations for all supported tools:
```
mr-archano: I'm not sure I understand why `necessary` is used here. Shouldn't be something like: "If you want to use the configuration from the artifact you can configure the tools as follows" and then I would list each tool in a different paragraph, as each configuration is independent from each other.

docs/advanced-usage.md
```diff
@@ -131,3 +132,57 @@ staticAnalysis {
...
+
+### Consume rules from artefact     
+
+In order to reuse your rules among multiple projects or to easily use an open source rule set, we added support for consuming the 
+rules for all supported tools from a maven artefact. An example of how to create and publish your rules as artefact can be 
+found [here](https://github.com/novoda/novoda/blob/master/scaffolding/build.gradle). Novoda open sourced their rules to [bintray](https://bintray.com/novoda/maven/static-analysis-rules) including
+configurations for Checkstyle, PMD, FindBugs, detekt and Android Lint.
+
+
+To consume the rules from an artefact you need to configure the `rules` closure pointing to a maven artefact and use it when configuring the tools, e.g.
```
mr-archano: This is not very clear to me. We need to define:
1 - what a rule artifact is: it's just a bundle of files that is published on a maven repository as artifact.
2 - how you define a dependency from a rule artifact: in order to access the files inside a rule artifact you have to first define an entry in the `rules {}` extension of the plugin. An entry is defined by a name and a maven coordinate, as follows:
```
staticAnalysis {
    rules {
        novoda {
            maven 'com.novoda:static-analysis-rules:0.2'
        }
    }
}
```
3 - once you have defined a rule artifact you can access the files inside it specifying the path of the file inside the bundle, eg:
```
def modules = rules.novoda['checkstyle-modules.xml']
```
Note that `modules` is defined as [`TextResource`](https://docs.gradle.org/current/dsl/org.gradle.api.resources.TextResource.html), that is a read-only body of text backed by a string, file, archive entry, or other source. Some of the tools already accepts `TextResource` as value for some of their configuration, while in other cases you have to transform a `TextResource` into a `File` or a path. Some examples of using a rule artifact in the supported tools is provided below:
- checkstyle sample
- pmd sample
- findbugs sample
- detekt
- lintoptions

docs/advanced-usage.md
rock3r: Both spellings go:

image

docs/advanced-usage.md
rock3r: `For reference, these are all the rules you can obtain from the novoda artefact:`

Although, I would rather have a table and only show what's relevant; e.g., have:

 Tool | Configuration | File name | Usage example
 --- | --- | --- | ---
 Checkstyle | `config` | `checkstyle-modules.xml` | `rules.novoda['checkstyle-modules.xml']`
 PMD | `ruleSetFiles` | `pmd-rules.xml` | `project.files(rules.novoda['pmd-rules.xml'].asFile().path)`

etc

rock3r:
Damn, it hadn't told me there were changes :/ I think most considerations are still valid although GH thinks they are not 💃

docs/advanced-usage.md
```diff
@@ -131,3 +132,57 @@ staticAnalysis {
 Please note that this is not yet supported for Detekt.
 
 [penaltyextensioncode]: https://github.com/novoda/gradle-static-analysis-plugin/blob/master/plugin/src/main/groovy/com/novoda/staticanalysis/PenaltyExtension.groovy
+
+
+### Consume rules from artefact
```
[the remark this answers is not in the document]
tobiasheine: Fixed 👍

docs/advanced-usage.md
tobiasheine: I prefer to have sth. that the user can easily copy & paste. The `toolsVersion` for example is also needed which I had to learn the hard way :(

docs/advanced-usage.md
```diff
@@ -131,3 +132,57 @@ staticAnalysis {
...
 
 [penaltyextensioncode]: https://github.com/novoda/gradle-static-analysis-plugin/blob/master/plugin/src/main/groovy/com/novoda/staticanalysis/PenaltyExtension.groovy
+
+
+### Consume rules from artefact     
+
+In order to reuse your rules among multiple projects or to easily use an open source rule set, we added support for consuming the 
+rules for all supported tools from a maven artefact. An example of how to create and publish your rules as artefact can be 
+found [here](https://github.com/novoda/novoda/blob/master/scaffolding/build.gradle). Novoda open sourced their rules to [bintray](https://bintray.com/novoda/maven/static-analysis-rules) including
+configurations for Checkstyle, PMD, FindBugs, detekt and Android Lint.
```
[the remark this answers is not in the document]
tobiasheine: That paragraph doesn't exist anymore

docs/advanced-usage.md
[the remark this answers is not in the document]
tobiasheine: That paragraph doesn't exist anymore

tobiasheine:
@rock3r comments are addressed 👍

tobiasheine:
@rock3r comments have been addressed 👍 let's 🚢 it

PR #101, opened 2018-03-15
Add multi-module sample project

takecare:
As setting up SAP on a multi-module project can be a bit tricky (or not very straight-forward) a sample project showcasing this sort of setup can be very helpful for people that want to use the plugin under such conditions.

The multi-module sample is a copy of the already existing sample, with an added (Java) module and the required changes to the SAP setup.


sample-multi-module/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+            jcenter()
+        }
+    }
+
+    repositories {
+        google()
+        mavenCentral()
+    }
+}
+
```
rock3r: Nit: extra whitelines

sample-multi-module/build.gradle
```diff
@@ -0,0 +1,32 @@
...
+    ext.kotlin_version = '1.2.30'
+
+    repositories {
+        google()
+        maven { url "https://plugins.gradle.org/m2/" }
+    }
+
+    dependencies {
+        classpath 'com.android.tools.build:gradle:3.0.1'
+        classpath "gradle.plugin.io.gitlab.arturbosch.detekt:detekt-gradle-plugin:1.0.0.RC6-3"
```
rock3r: RC6-4 is out FYI

sample-multi-module/team-props/tasks.gradle
```diff
@@ -0,0 +1,13 @@
...
+
+task check(dependsOn: subprojects.collect { p ->
+    "$p.name:check"
+})
+
+task assemble(dependsOn: subprojects.collect { p ->
+    "$p.name:assemble"
+})
+
+task build(dependsOn: ['assemble', 'check'])
```
rock3r: Are we sure we need these three tasks? Isn't it the default behaviour?

takecare: right. the way we refer to the tasks just needs some changing.

sample-multi-module/build.gradle
takecare: seems to be breaking the plugin 🙀

takecare: tracked in #102

PR #106, opened 2018-05-25
Gradle: Adding support for snapshots

tomsammons:
This PR adds support for snapshots builds

These will be built off the development branch 

We updated the `gradle-build-properties-plugin` to the latest version

Here is the link to the Jenkins job: https://ci.novoda.com/job/gradle-static-analysis-plugin-snapshot/

Paired with @tobiasheine

PR #107, opened 2018-05-28
Snapshot: Using correct repo 

tomsammons:
Previously had the repo hardcoded to use maven. Now we inherit this value from the environment.

### DIFF
--- a/gradle/publish.gradle
+++ b/gradle/publish.gradle
@@ -52,7 +52,7 @@ apply plugin: 'com.novoda.bintray-release'
 
 publish {
     userOrg = 'novoda'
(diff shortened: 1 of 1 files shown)


gradle/publish.gradle
```diff
@@ -52,7 +52,7 @@ apply plugin: 'com.novoda.bintray-release'
 
 publish {
     userOrg = 'novoda'
-    repoName = 'maven'
+    repoName = buildProperties.publish['bintrayRepo'].string
```
mr-archano: 🙈 sorry @tomsammons I completely missed this in #106

PR #109, opened 2018-07-03
Exclude kotlin files by default on java code quality tools.

### LINKED ISSUE #56: Exclude kotlin files for Checkstyle, PMD and Findbugs
As mentioned by @rock3r we don't want to run the built-in static analysis tools (Checkstyle, PMD, FindBugs) on Kotlin files as we have dedicated tools for them.
To streamline the usage of the plugin we could just add default `exclude` rules to each of the tools above.

tasomaniac:
This PR adds a default exclusion for kotlin files on Checkstyle, Findbugs and PMD tools.

Note: Since `sample` is in master, I couldn't change it to reflect these changes. That needs to be updated separately. 

~Also update samples to remove redundant excludes.~

~Also fixes the detekt config not to fail so that the plugin handles violations. This is already the recommended approach in [our guides](https://github.com/tasomaniac/gradle-static-analysis-plugin/blob/master/docs/tools/detekt.md#configure-detekt). Sample should reflect that.~

Fixes #56

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
@@ -83,7 +83,10 @@ abstract class CodeQualityConfigurator<T extends SourceTask, E extends CodeQuali
     protected abstract void configureAndroidVariant(variant)
 
     protected void configureJavaProject() {
-        project.tasks.withType(taskClass) { task -> sourceFilter.applyTo(task) }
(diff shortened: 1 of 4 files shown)


hal9002:
Can one of the admins review this PR?

rock3r:
hal test this please

rock3r:
Oh good catch, totally missed that myself

tasomaniac:
Aha! 💡Will do.

tasomaniac:
@mr-archano @rock3r please take another look :)

rock3r:
CI please test this

rock3r:
hal test this please

PR #118, opened 2018-08-01
Merge master back to develop for sample

tasomaniac:
I want to do some docs for ktlint but the sample is not in develop yet. So I thought it would be good to bring changes in master into develop


tasomaniac:
Bump! This may need manual merge.

PR #121, opened 2018-08-11
Fix for newer Detekt versions

### LINKED ISSUE #102: Detekt configuration fails with RC6-4
Breaking changes with https://github.com/arturbosch/detekt/pull/719. `systemOrDefaultProfile()` is no longer available.

```
* Where:
Script '.../gradle-static-analysis-plugin/sample-multi-module/team-props/static-analysis.gradle' line: 36

* What went wrong:
A problem occurred evaluating script.
> Could not find method systemOrDefaultProfile() for arguments [] on DetektExtension(version='latest.release', debug=false, profile='main', ideaExtension=IdeaExtension(path=null, codeStyleScheme=null, inspectionsProfile=null, report=null, mask='*.kt'), profiles=[ProfileExtension(name='main', input=.../gradle-static-analysis-plugin/sample-multi-module/app/src/main/java, config=.../gradle-static-analysis-plugin/sample-multi-module/team-props/detekt-config.yml, configResource=/default-detekt-config.yml, filters=.*test.*,.*/resources/.*,.*/tmp/.*, ruleSets=null, output=.../gradle-static-analysis-plugin/sample-multi-module/app/build/reports/detekt, outputName=null, baseline=null, parallel=false, disableDefaultRuleSets=false)]) of type io.gitlab.arturbosch.detekt.extensions.DetektExtension.
```

tasomaniac:
## Problem

- `Profile` public property in Detekt extension was moved in `1.0.0-RC6.3`
  - This caused `systemOrDefaultProfile()` method call to fail as it is not available anymore. 

## Solution

Check if a property or method is available before calling it. 
If not available, throw a message. Feedback is welcome on the message itself. 

Also checked the whole integration. The rest seems to be fine. Only other thing that we access is the actual Gradle task name which is not likely to change. 

Also added tests to cover both Detekt versions.

Fixes #102

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
@@ -13,6 +13,7 @@ class DetektConfigurator implements Configurator {
     private static final String DETEKT_PLUGIN = 'io.gitlab.arturbosch.detekt'
     private static final String DETEKT_NOT_APPLIED = 'The Detekt plugin is configured but not applied. Please apply the plugin in your build script.\nFor more information see https://github.com/arturbosch/detekt.'
     private static final String OUTPUT_NOT_DEFINED = 'Output not defined! To analyze the results, `output` needs to be defined in Detekt profile.'
+    private static final String DETEKT_CONFIGURATION_ERROR = 'A problem occurred while configuring Detekt. Please make sure to use a compatible version.'
 
     private final Project project
     private final Violations violations
@@ -53,9 +54,6 @@ class DetektConfigurator implements Configurator {
         def detektTask = project.tasks['detektCheck']
         detektTask.group = 'verification'
 
-        // run detekt as part of check
-        project.tasks['check'].dependsOn(detektTask)
-
         // evaluate violations after detekt
         def output = resolveOutput(detekt)
         if (!output) {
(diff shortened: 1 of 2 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
```diff
@@ -53,11 +54,8 @@ class DetektConfigurator implements Configurator {
         def detektTask = project.tasks['detektCheck']
         detektTask.group = 'verification'
 
-        // run detekt as part of check
-        project.tasks['check'].dependsOn(detektTask)
```
tasomaniac: This already transitively happens since `check` dependsOn `evaluateViolations`. No other tool configurator does this. Removed!

plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy
```diff
@@ -13,6 +13,7 @@ class DetektConfigurator implements Configurator {
     private static final String DETEKT_PLUGIN = 'io.gitlab.arturbosch.detekt'
     private static final String DETEKT_NOT_APPLIED = 'The Detekt plugin is configured but not applied. Please apply the plugin in your build script.\nFor more information see https://github.com/arturbosch/detekt.'
     private static final String OUTPUT_NOT_DEFINED = 'Output not defined! To analyze the results, `output` needs to be defined in Detekt profile.'
+    private static final String DETEKT_CONFIGURATION_ERROR = 'A problem occurred while configuring Detekt. Please make sure to use a compatible version.'
```
rock3r: Would be nice to write what versions are compatible

tasomaniac: Yes, very good point. I was thinking about it but I couldn't think about a way to do it in a maintainable way. It is at least compatible up to `RC8`. 

Considering this is shown only in case of a future incompatibility, it should be fine.

tasomaniac:
Thanks for the review @rock3r Please have another look.

tasomaniac:
Yes, we are just trying to access the reports. They will be available to us just like the other tools. :+1:

PR #124, opened 2018-08-13
Update docs for ktlint integration

tasomaniac:


docs/tools/ktlint.md
```diff
@@ -0,0 +1,58 @@
...
+```gradle
+ktlint {
+    android true
+}
+```
+
+For other configuration options and adding custom rules, refer to the 
+[official guide](https://github.com/JLLeitschuh/ktlint-gradle/#configuration).
+
+**Note:** Please make sure __note__ to set `ignoreFailures = false` since failures and threshold detection is handled by
```
rock3r: I think that `__note__` in the middle is a leftover of something else, and to double check: I would expect `ignoreFailures` to be set to `true` because we don't want ktlint to break the build itself

PR #128, opened 2018-08-25
Update dead links in supported tools docs

rock3r:
As it says on the tin


tasomaniac:
LGTM. I wonder if there is a way to detect broken links

rock3r:
I know [Flutter.io](https://github.com/flutter/website#preventing-broken-links) does have something like that but not sure how easy it'd be to port over

PR #130, opened 2018-09-05
PT-821: Downgrade Gradle

tobiasheine:
Tracked in JIRA as [PT-821](https://novoda.atlassian.net/browse/PT-821)

**Description**
Upgrading Gradle from `4.4.1` to `4.9` broke the release.

**Considerations**
As quick-fix, we will downgrade Gradle back to `4.4.1`.

**Tests**
Verified manually using the `dry-run` property that all fields are set correct to the `publish` closure.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.9-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4.1-all.zip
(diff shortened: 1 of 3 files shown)


rock3r:
We have a bunch of broken tests, was Gradle not the problem then?

tobiasheine:
@rock3r 
Nope, we had failing tests cause we updated the android SDK on the CI and lint kicked in with `OldTargetApi`. I disabled this check for the tests.

tasomaniac:
Would be good to know more about the problem. I'm using the plugin in production together with Gradle 4.9 and 4.10

tobiasheine:
@tasomaniac The issue is related to the [bintray release plugin](https://github.com/novoda/bintray-release/issues/220), which doesn't work with Gradle 4.9 or a newer version.

PR #132, opened 2018-09-24
Fix dead links in supported-tools.md

rock3r:
Does what it says on the tin


rock3r:
Tests are fixed in #133, it's an unrelated issue.

PR #133, opened 2018-09-24
Pin Detekt version in tests

rock3r:
Tests are going red because we weren't pinning the Detekt version in the tests. This fixes it.


tasomaniac:
I would expect Detekt plugin to use the plugin version by default. Interesting that they always use the latest version. The results are not reproducible unless you pin the version manually. Interesting. 😊

rock3r:
It does use the corresponding version by default only starting in RC9, before it was the latest available

PR #141, opened 2018-10-08
Update samples

mr-archano:
This PR updates the samples to use the new Detekt Gradle Plugin (`1.0.0.RC9`).
Incidentally I have used composite builds to consume a local build of the plugin instead of relying on the `buidSrc` trick.


sample-multi-module/team-props/static-analysis.gradle
```diff
@@ -34,11 +34,12 @@ staticAnalysis {
...
-        profile('main') {
-            input = project.file("src/main/java")
-            config = rootProject.file('team-props/detekt-config.yml')
-            filters = '.*test.*,.*/resources/.*,.*/tmp/.*'
-            output = project.file("build/reports/detekt")
+        input = project.files("src/main/java")
+        config = rootProject.files('team-props/detekt-config.yml')
+        filters = '.*test.*,.*/resources/.*,.*/tmp/.*'
+        reports {
+            xml.enabled = true
```
mr-archano: @tasomaniac I was wondering: shouldn't we add this in `DetektConfigurator`? We actually switch both xml and html reports on in other tools too.

Issue #140, opened 2018-10-08
KtLint 6.x incompatibility


tasomaniac:
Unfortunately KtLint 6.0.0 has a behavior change. It is pretty minimal but still breaks our plugin 😒

The reporters config has changed, instead of accepting strings, it requires their internal enum type.

That could be a problem for us. In that case, we could remove setting reporter and let the user do their reporting setup.

Issue #129, opened 2018-09-03
KtLintConfigurator fails to find tasks


tasomaniac:
Just tried to integrate it into our project. Unfortunately SAP is failing to resolve the `ktlint` tasks even thought tasks are actually present. I think it is a timing issue. 

```
Caused by: org.gradle.api.UnknownTaskException: Task with name 'ktlintWayfairTangoDisabledDebugAndroidTestCheck' not found in project ':app'.
        at org.gradle.api.internal.tasks.DefaultTaskCollection.createNotFoundException(DefaultTaskCollection.java:88)
        at org.gradle.api.internal.tasks.DefaultTaskCollection.createNotFoundException(DefaultTaskCollection.java:39)
        at org.gradle.api.internal.DefaultNamedDomainObjectCollection.getByName(DefaultNamedDomainObjectCollection.java:320)
        at org.gradle.api.internal.DefaultNamedDomainObjectCollection.getAt(DefaultNamedDomainObjectCollection.java:339)
        at org.gradle.api.internal.tasks.DefaultTaskCollection.getAt(DefaultTaskCollection.java:39)
        at com.novoda.staticanalysis.internal.ktlint.KtlintConfigurator$_createCollectViolationsTask_closure6.doCall(KtlintConfigurator.groovy:94)
```

rock3r:
Is the KtLint plugin using `afterEvaluate`? Because that would explain the issue.

I have [set it up on Squanchy](https://github.com/squanchy-dev/squanchy-android/pull/585) tho and it works fine.

tasomaniac:
It was also working fine for me for couple of simpler personal projects. I didn't investigate further but I'm suspecting a timing issue resulting because of the size of Wayfair app.

I've linked an issue I opened in KtLint. It is probably because of afterEvaluate

mr-archano:
Dammit. Nested `afterEvaluate` are PITA. I wonder if there's anything else we can do to prevent the issue altogether - a callback similar to `plugins.withId` maybe, like `tasks.withType()`? Not sure...

tasomaniac:
I am solving for now in our project by adding another afterEvaluate lol 😀

tasomaniac:
Actually this is supposed to be fixed with Ktlint plugin version 6.x and later. But #140 prevents me to verify that.

PR #145, opened 2018-11-08
Fix some ktlint issues

tasomaniac:
Tried to fix some ktlint issues. 

- Added tests for different versions
- Automatically resolve the output files name from the task. File names have been changed but not this property. 
- Removed automatic adding of `CHECKSTYLE` reporter since passing String is not supported anymore and we cannot really reference their enum type in our integration. 
  - That's why now, we will require users to manually add `CHECKSTYLE` in their config.
  - Our integration will crash with a good message if it is missing. 

Problems: 
- Only for Android projects, for some reason `ktlintCheck` task give `NO-SOURCE` no matter I tried. Will spend some more time later.

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
@@ -61,6 +61,9 @@ class KtlintConfigurator implements Configurator {
                 project.plugins.withId("kotlin-platform-common") {
                     configureKotlinProject()
                 }
+                project.plugins.withId("org.jetbrains.kotlin.multiplatform") {
+                    configureKotlinProject()
+                }
                 project.plugins.withId('com.android.application') {
                     configureAndroidWithVariants(variantFilter.filteredApplicationVariants)
                 }
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
@@ -9,11 +9,13 @@ import org.gradle.api.GradleException
 import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.Project
 import org.gradle.api.Task
+import org.gradle.api.file.RegularFileProperty
 
 class KtlintConfigurator implements Configurator {
 
     private static final String KTLINT_PLUGIN = 'org.jlleitschuh.gradle.ktlint'
     private static final String KTLINT_NOT_APPLIED = 'The Ktlint plugin is configured but not applied. Please apply the plugin in your build script.\nFor more information see https://github.com/JLLeitschuh/ktlint-gradle/#how-to-use'
+    private static final String XML_REPORT_NOT_ENABLED = 'XML report must be enabled. Please make sure to add "CHECKSTYLE" into reports in your Ktlint configuration'
 
     private final Project project
     private final Violations violations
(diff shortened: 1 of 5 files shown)


tasomaniac:
Fixes #140

tasomaniac:
Also fixes #129 since I use `matching` to resolve the task dynamically.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
```diff
@@ -9,11 +9,13 @@ import org.gradle.api.GradleException
 import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.Project
 import org.gradle.api.Task
+import org.gradle.api.file.RegularFileProperty
```
mr-archano: Is this putting a cap on the minimum version of Gradle we support now?

plugin/src/test/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintIntegrationTest.groovy
```diff
@@ -73,22 +97,22 @@ class KtlintIntegrationTest {
...
+                result.buildFileUrl("reports/ktlint/$expectedOutputFileName"))
     }
 
     @Test
     void shouldNotFailBuildWhenNoErrorsEncounteredAndNoThresholdTrespassed() {
         def result = createProjectWith(Fixtures.Ktlint.SOURCES_NO_ERROR, 0)
-                .withToolsConfig(EMPTY_CONFIG)
-                .build('evaluateViolations')
+                .withToolsConfig(DEFAULT_CONFIG)
+                .build('evaluateViolations', '-s')
```
mr-archano: I wonder if we should add `-s` to the default parameters instead 🤔

plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
mr-archano: I can see from the javadocs that this is since Gradle 4.3. I don't think this is a blocker per-se, just never thought of the implicit compatibility with Gradle API until now.

tasomaniac: Hmmm, since ktlint requires Gradle 4.3 anyways, it might not be a problem. I wonder what happens with the other tools when Gradle is old and ktlint is not used. I will check that.

plugin/src/test/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintIntegrationTest.groovy
tasomaniac: That's a mistake :)

plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
mr-archano: Probably not worth the hassle right now. We might have already an implicit dependency we never realised before. Also worst case scenario this will crash on projects running on Gradle < 4.3 only at runtime (and as you said KtLint will need 4.3+ anyway).

tasomaniac: If you're sure that it is a runtime crash, then I think we are good to go with that. :+1:

mr-archano: I'm like 99.9% sure that will be a crash at configuration time - the same it will happen when trying to use KtLint with older versions of Gradle

tasomaniac: It will crash at configuration only when KtLint is used right?

tasomaniac: I tried it but not so easy in a single project. Will try once we get a snapshot.

plugin/src/test/groovy/com/novoda/test/TestAndroidKotlinProject.groovy
```diff
@@ -26,7 +26,6 @@ apply plugin: 'kotlin-android'
 
 android {
     compileSdkVersion 27
-    buildToolsVersion '27.0.0'
```
tasomaniac: These were giving warnings. Not really needed.

PR #153, opened 2018-12-03
Fix ktlint for all versions

tasomaniac:
Ktlint integration is fixed for 6.x  and upcoming 7.x versions. 

For Ktlint integration, we only depend on task names and a certain property (`reportOutputFiles`). Extracted task `matching` into a function and support resolving the tasks from all versions of ktlint.

## Tests

~6.1.0 is also fixed but has a bug affecting our test setup. I don't think it is worth to fix that only for tests. 6.2.x and 6.3.x are supported.~

Also enabled tests for `6.1.0` version by allowing test fixtures to be copied into the actual test project.

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
@@ -97,12 +97,19 @@ class KtlintConfigurator implements Configurator {
 
     private void configureKtlint(String sourceSetName) {
         project.tasks.matching {
-            it.name == "ktlint${sourceSetName.capitalize()}Check"
+            isKtlintTask(it, sourceSetName.capitalize())
         }.all { Task ktlintTask ->
             def collectViolations = configureKtlintWithOutputFiles(sourceSetName, ktlintTask.reportOutputFiles)
             collectViolations.dependsOn ktlintTask
             evaluateViolations.dependsOn collectViolations
         }
+    }
+
+    /**
+     * KtLint task has the following naming convention and the needed property to resolve the reportOutputFiles
+     */
+    private boolean isKtlintTask(Task task, String sourceSetName) {
+        task.name ==~ /^ktlint$sourceSetName(SourceSet)?Check$/ && task.hasProperty('reportOutputFiles')
     }
 
     private def configureKtlintWithOutputFiles(String sourceSetName, Map<?, RegularFileProperty> reportOutputFiles) {
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
@@ -93,7 +93,6 @@ class KtlintConfigurator implements Configurator {
         variant.sourceSets.each { sourceSet ->
             configureKtlint(sourceSet.name)
         }
-        configureKtlint(variant.name)
     }
 
     private void configureKtlint(String sourceSetName) {
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
@@ -84,12 +84,19 @@ class KtlintConfigurator implements Configurator {
     }
 
     private void configureAndroidWithVariants(def mainVariants) {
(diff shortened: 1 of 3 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
```diff
@@ -84,12 +84,19 @@ class KtlintConfigurator implements Configurator {
...
+        variantFilter.filteredTestVariants.all { configureAndroidVariant(it) }
+        variantFilter.filteredUnitTestVariants.all { configureAndroidVariant(it) }
     }
 
-    private void configureKtlint(def sourceSetName) {
+    private void configureAndroidVariant(def variant) {
+        variant.sourceSets.each { sourceSet ->
+            configureKtlint(sourceSet.name)
+        }
+        configureKtlint(variant.name)
```
mr-archano: isn't this already covered by L93-L95? also, is it possible to use `.all {}` here instead?

mr-archano: isn't this already covered by L93-L95? also, is it possible to use `.all {}` here instead?

plugin/src/test/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintIntegrationTest.groovy
```diff
@@ -35,12 +35,12 @@ class KtlintIntegrationTest {
                 [TestProjectRule.forKotlinProject(), '5.1.0', 'ktlint-main.txt'],
                 [TestProjectRule.forAndroidKotlinProject(), '5.1.0', 'ktlint-debug.txt'],
                 [TestProjectRule.forKotlinProject(), '6.1.0', 'ktlintMainCheck.txt'],
+                // Fails for our test setup since we have custom sourceDirs. https://github.com/JLLeitschuh/ktlint-gradle/issues/153
```
mr-archano: I'm not sure I understand why this is happening 🤔

plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
tasomaniac: Don't know exactly why but KtLint plugin changed from supporting variants to sourceSets. That's why it was broken. So this two together kinda support all versions automatically. 

On a second thought. Maybe it is not necessary. I will try without it.

tasomaniac: I tried that but didn't work. 

Since we use variants.all, at this point, variant is resolved with all details. At this point sourceSets is not dynamic and does not have .all method.

plugin/src/test/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintIntegrationTest.groovy
tasomaniac: They had a bug where they resolved sourceSets too early. And additional source dirs were not checked.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
tasomaniac: Fixed in https://github.com/novoda/gradle-static-analysis-plugin/pull/153/commits/be0bf8269351081ffcd8c184d258bc61508d6d34

plugin/src/test/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintIntegrationTest.groovy
tasomaniac: Enabled this test now.

tasomaniac:
Hey @mr-archano, did some changes to support the upcoming ktlint changes. I think this is ready to review and merge. Once they publish the plugin, I will add new tests for `7.x` versions.

Also had a small addition in test to enable the commented out test.

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
```diff
@@ -63,6 +63,13 @@ ${project.additionalConfiguration}
         file.text = text
     }
 
+    public T copyIntoSourceSet(String sourceSet, File srcDir) {
```
mr-archano: 💯

tasomaniac:
retest please

tasomaniac:
@hal90002 retest this please

PR #154, opened 2018-12-05
Make Findbugs Html report optional

### LINKED ISSUE #117: Allow to disable Findbugs HTML reports creation
As mentioned in https://github.com/novoda/gradle-static-analysis-plugin/issues/74, every time we run `check` a `GenerateFindBugsHtmlReport` task is created and run.
While this is good when running the plugin on a dev machine, it's just wasting time when it's running on a CI where there are other ways to display the checks report (for example using the Pull Request Builder plugin).

It would be good to have an option to enable/disable the html report directly in the findbugs configuration closure.

tasomaniac:
Added `htmlReportEnabled` to the DSL to control whether we should generate html report or not. 

Initially I wanted to use `reports.html.enabled` but that is not available. `FindbugsExtension` is pretty limited and unfortunately does not have that built-in. It is in `Findbugs` task and to be able to use that we would require users to use something like `tasks.withType(Findbugs) { reports.html.enabled = false }` which is ugly.

Fixes #117

### DIFF
--- a/docs/tools/findbugs.md
+++ b/docs/tools/findbugs.md
@@ -17,6 +17,7 @@ findbugs {
     toolVersion // A string, most likely '3.0.1' — the latest Findbugs release (for a long time)
     exclude // A fileTree, such as project.fileTree('src/test/java') to exclude Java unit tests
     excludeFilter // A file containing the Findbugs exclusions, e.g., teamPropsFile('static-analysis/findbugs-excludes.xml')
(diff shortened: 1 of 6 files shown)


mr-archano:
@hal90002 retest this please

tasomaniac:
Just updated it 😄Let's see what CI says now :)

Btw, I did couple of commits and reverts. Feel free to squash and merge.

PR #157, opened 2018-12-10
Fix for files not being in the srcDirs

tasomaniac:
Right now, we put the following into Test Project setup to create the test projects:

```
sourceSets { 
    main {
        srcDir full/path/to/folder/with/fixtures
    }
}
```

This folder is outside of the created test project. In real projects, this rarely happens. And this has caused problems couple of times with the tests (in some instances they're actual bugs in the tools).

With this PR, we now copy the whole file(s) into respective `sourceSet` folder.


tasomaniac:
Hmmm, missed the nested folders. Will close this for now.

Issue #19, opened 2017-01-31
Show total count of issues found


xrigau:
Every time I go on a boyscout mission and fix some static analysis issues I end up with the number of issues for each analysis type, but I have to configure a threshold for the total number of issues:

```
> Checkstyle rule violations were found (58 errors, 43 warnings). See the reports at:
- file:///Users/xrigau/development/git/novoda/some-cool-project/app/build/reports/checkstyle/main.html
> Findbugs rule violations were found (9 errors, 6 warnings). See the reports at:
- file:///Users/xrigau/development/git/novoda/some-cool-project/app/build/reports/findbugs/coolprojectStaging.html
> PMD rule violations were found (11 errors, 13 warnings). See the reports at:
- file:///Users/xrigau/development/git/novoda/some-cool-project/app/build/reports/pmd/main.html
```

It'd be a lot easier if it printed a line such as:

```
Total: 78 errors, 62 warnings
```

So that I can update the thresholds after every boyscoutting more easily.

This wouldn't be needed if #18 was implemented

mr-archano:
This can be easily added indipendently from #18. Let's do it 👍

tasomaniac:
Regarding this, I was also thinking about merging results across sub-projects in a big Gradle project. 
Imagine a project with many modules, when a tool fails the build, it is hard to detekt which module it is. And reports are separated since they run individually on all projects.

Also in such scenario, `maxErrors` does not make much sense individually for each submodule. It would only makes sense when `maxErrors = 0` (simply because  `0 x noOfModules = 0`)

In past projects, I did the following to merge unitTest results. This is something provided by Gradle. I couldn't find an easy way to do the same for `lint` for example (i.e merging lint results from different modules)

Merging multiple unit test results:
```
  def testTasks = [...] // manual list of all test tasks

  task testReport(type: TestReport) {
    destinationDir = file("$buildDir/reports/allTests")
    reportOn { testTasks.collect { project.tasks.findByPath(it)*.binResultsDir } }
  }
```

tasomaniac:
@xrigau today is your lucky day

xrigau:
EXCITEMENT!

![](http://forgifs.com/gallery/d/213102-1/Baby-omg-reactions.gif)

tasomaniac:
I thought you would have 0 warnings and error by now :troll:

xrigau:
![](https://i.imgur.com/XgnfHrr.gif)

PR #159, opened 2018-12-11
Small adjustments to the output

### LINKED ISSUE #155: Empty output when everything is alright
`collect${toolName}Violations` and `evaluateVialoations` tasks output empty space whenever the task is successful. By default, the newest (non-plain) Gradle console just skips the empty output tasks. It is a really nice feature. But empty lines from `evaluateViolations` task break it. Its more of a problem if you have many Gradle subprojects. 

TODO: 
If there is nothing to output skip it. It could be as simple as calling `trim()` on the output or actually removing empty lines.

image

tasomaniac:
Fix empty space problem with the output. Verified that successful builds/checks wouldn't output anything at all on a multi-module environment.

Also fixes an old issue #19 where we put the total number of issues even when the task is successful. Will help getting an idea when there are too many warnings.

Also modified the tests to test the whole message.

Fixes #155 #19

### DIFF
--- a/plugin/src/test/groovy/com/novoda/test/LogsSubject.groovy
+++ b/plugin/src/test/groovy/com/novoda/test/LogsSubject.groovy
@@ -44,10 +44,6 @@ class LogsSubject extends Subject<LogsSubject, Logs> {
         check().that(actual().output)
     }
 
-    public void isEmpty() {
-        outputSubject.isEmpty()
-    }
-
     public void contains(log) {
         outputSubject.contains(log)
     }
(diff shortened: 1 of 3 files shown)


plugin/src/test/groovy/com/novoda/test/LogsSubject.groovy
```diff
@@ -44,6 +44,10 @@ class LogsSubject extends Subject<LogsSubject, Logs> {
         check().that(actual().output)
     }
 
+    public void isEmpty() {
```
mr-archano: `nit:` this is new but doesn't look like it's used anywhere in the PR

tasomaniac: Oh, I used it and then replaced it with verifyZeroInteractions

Issue #33, opened 2017-04-19
Noisy logging


ouchadam:
Running `./gradlew evaulateViolations` produces lots of unnecessary logs, is it possible to turn them off? 

```
Scanning archives (0 / 221)
Scanning archives (1 / 221)
Scanning archives (2 / 221)
Scanning archives (3 / 221)
Scanning archives (4 / 221)
Scanning archives (5 / 221)
Scanning archives (6 / 221)
Scanning archives (7 / 221)
Scanning archives (8 / 221)
Scanning archives (9 / 221)
Scanning archives (10 / 221)
...

Pass 1: Analyzing classes (0 / 520) - 00% complete
Pass 1: Analyzing classes (1 / 520) - 00% complete
Pass 1: Analyzing classes (2 / 520) - 00% complete
Pass 1: Analyzing classes (3 / 520) - 00% complete
Pass 1: Analyzing classes (4 / 520) - 00% complete
Pass 1: Analyzing classes (5 / 520) - 00% complete
Pass 1: Analyzing classes (6 / 520) - 01% complete
Pass 1: Analyzing classes (7 / 520) - 01% complete
Pass 1: Analyzing classes (8 / 520) - 01% complete
Pass 1: Analyzing classes (9 / 520) - 01% complete
...
```

mr-archano:
Apparently something changed in the newer versions of Gradle, and now the static analysis tools are producing more logs than before. In my experience this started happening since Gradle 3.x, but we need to investigate whether we can turn them off from the plugins. Setting explicitly a quiet logger as it's done now apparently is not enough 👯

tasomaniac:
Together with #160 this is really not an issue anymore. Closing now.

PR #160, opened 2018-12-11
Less verbose Findbugs output

tasomaniac:
I was looking at #33 and realized that I've been seeing Android classes in the output of Findbugs.

It turns out that these are called `auxclass` since they are not available to findbugs. Adding `auxclass` parameter removes these warnings and the output becomes less verbose

Before | After
------ | ------
| image | image |


plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
```diff
@@ -67,6 +67,7 @@ class FindbugsConfigurator extends CodeQualityConfigurator<FindBugs, FindBugsExt
             description = "Run FindBugs analysis for ${variant.name} classes"
             source = androidSourceDirs
             classpath = variant.javaCompile.classpath
+            extraArgs '-auxclasspath', androidJar
```
mr-archano: 😮

PR #162, opened 2018-12-18
Prepare for Release 0.8

tasomaniac:


build.gradle
```diff
@@ -29,7 +29,7 @@ subprojects {
     }
 }
 
-task wrapper(type: Wrapper) {
+wrapper {
```
tasomaniac: this is the new way. 👍 And the old one is already forbidden in Gradle 5.0

tasomaniac:
The task just prints it with backticks. I guess that's fine.

PR #165, opened 2019-01-14
WIP: Task Configuration Avoidance

tasomaniac:
https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

This is a feature that became stable in Gradle 5.1 but the APIs were introduced in Gradle 4.9

This allows lazy configuration of tasks. Android Tools already migrated in 3.3 and 3.4 and saved configuration of thousands of tasks. Unfortunately our plugin resolves those tasks. Since they are the entry points of the whole system, that transitively configures thousands more. 

This is a WIP that allows nothing to be configured unless `evaluateViolations` is run.

Running `evaluateViolations` explicitly will have 0 behavior change. This is how the plugin is supposed to be used, so it is no problem. But still I think I need to explain the rest. When `evaluateViolations` is not run, task dependencies are not necessarily set since they are lazy and may not be run/configured. Some of the `collect${toolName}Violations` tasks may not even be available anymore unless `evaluateViolations` is run.

**Disclamier:** Currently opened the PR for pre-review and testing with CI.

I need to do more intensive manual testing to make sure we don't configure any task eagerly.

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisPlugin.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisPlugin.groovy
@@ -1,6 +1,7 @@
 package com.novoda.staticanalysis
 
-import com.novoda.staticanalysis.internal.CodeQualityConfigurator
+import com.novoda.staticanalysis.internal.CollectViolationsTask
+import com.novoda.staticanalysis.internal.Configurator
 import com.novoda.staticanalysis.internal.checkstyle.CheckstyleConfigurator
 import com.novoda.staticanalysis.internal.detekt.DetektConfigurator
 import com.novoda.staticanalysis.internal.findbugs.FindbugsConfigurator
@@ -10,7 +11,8 @@ import com.novoda.staticanalysis.internal.pmd.PmdConfigurator
 import org.gradle.api.NamedDomainObjectContainer
 import org.gradle.api.Plugin
 import org.gradle.api.Project
-import org.gradle.api.Task
+
+import static com.novoda.staticanalysis.internal.TasksCompat.createTask
 
 class StaticAnalysisPlugin implements Plugin<Project> {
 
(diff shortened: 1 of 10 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/StaticAnalysisPlugin.groovy
```diff
@@ -10,39 +11,45 @@ import com.novoda.staticanalysis.internal.pmd.PmdConfigurator
...
-        project.tasks.create('evaluateViolations', EvaluateViolationsTask) { task ->
+    private static def createEvaluateViolationsTask(Project project, StaticAnalysisExtension extension) {
+        createTask(project, 'evaluateViolations', EvaluateViolationsTask) { task ->
             task.evaluator = { extension.evaluator }
             task.allViolations = { extension.allViolations }
+
+            project.tasks.withType(ToolTriggerTask) { toolTrigger ->
+                task.dependsOn(toolTrigger)
+            }
+            project.tasks.withType(CollectViolationsTask) { collectViolations ->
```
tasomaniac: `dependsOn` relation is now automatically added for all types of `CollectViolationsTask`s. No need to pass `evaluateViolations` anymore. 

But I have to put it into configuration of `evaluateViolations` which means this relation is lazily put once `evaluateViolations` is run.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/TasksCompat.groovy
```diff
@@ -0,0 +1,44 @@
...
+import org.gradle.util.GradleVersion
+
+class TasksCompat {
+
+    private static boolean IS_GRADLE_MIN_49 = GradleVersion.current() >= GradleVersion.version("4.9")
+
+    static <T extends Task> Object maybeCreateTask(Project project, String name, Class<T> type, Action<? super T> configuration) {
+        if (IS_GRADLE_MIN_49) {
+            try {
+                def provider = project.tasks.named(name)
```
tasomaniac: Here is an ugly solution to lack of `maybeRegister` API in Gradle. More info: https://github.com/gradle/gradle/issues/6243

Using `maybeCreate` may not be the best practice after all. But I've looked at our implementation. Since we want to be flexible and adjust to configuration of other tools and plugins, it is not easy to get rid of it. It is really valuable for us. 

They don't have lazy alternative of `maybeCreate`. But they also don't have any `lazy` query method. `named` is the new way. It returns `TaskProvider` instead. But it throws exception when the task is not found.

So, long story short, this is the only way to do lazy and maybe together.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
```diff
@@ -46,29 +44,34 @@ class KtlintConfigurator implements Configurator {
...
-                }
-                project.plugins.withId("org.jetbrains.kotlin.multiplatform") {
-                    configureKotlinProject()
-                }
-                project.plugins.withId('com.android.application') {
-                    configureAndroidWithVariants(variantFilter.filteredApplicationVariants)
-                }
-                project.plugins.withId('com.android.library') {
-                    configureAndroidWithVariants(variantFilter.filteredLibraryVariants)
+                createTask(project, 'triggerKtlint', ToolTriggerTask) { trigger ->
```
tasomaniac: Our `ktlint` integration is the most special. We dynamically look at the live collection of tasks and pick up ktlint tasks and configure them later. And we create `collectViolations` tasks based on availability to `ktlint` tasks. When we want to keep all of these lazy, it is simply a deadlock.

The only solution that I can come up with is to introduce another task called `trigger`. Calling this will resolve all ktlint tasks with no problem. When this task is called, we are sure that it is safe to configure everything.

tasomaniac:
Closing this for now. I will open again with smaller MR to introduce lazy tasks as much as possible. And will have 2nd iteration to fully make them lazy.

PR #167, opened 2019-01-31
Included new version of Ktlint plugin into tests

tasomaniac:
Just added the new versions to verify that everything is working as expected.


tasomaniac:
@mr-archano can we also merge this one? It's a little test.

Issue #168, opened 2019-02-02
Upgrade Gradle to 5.1.1


zegnus:
The current used version is 4.10.2 https://github.com/novoda/gradle-static-analysis-plugin/blob/master/sample-multi-module/gradle/wrapper/gradle-wrapper.properties#L3

It would be great to ensure that the plugin supports the latest 5.x available Gradle version, including the sample applications.

tasomaniac:
@zegnus, I can verify that it works fine with Gradle 5.x. 

Since we don't use any of internal APIs, it should always be backward compatible. 

I have an open PR #165 where I try to integrate lazy task configuration from Gradle 5 in a backward compatible fashion. We will definitely upgrade to 5.1.1 together with that. 

There is no practicle reason to update now. Having said that, feel free to open a PR for the upgrade. 🙌

zegnus:
I cannot open a PR for this upgrade because the sample applications fail:

In `sample/app/gradle/wrapper/gradle-wrapper.properties` use `gradle-5.1.1-all.zip`

then, inside the `sample` directory, run:

`./gradlew checkstyleMain`

It fails with the following:
```
Execution failed for task ':app:checkstyleMain'.
> Cannot convert org.gradle.api.internal.file.collections.DefaultSingletonFileTree@3e0a3cdc to local file system directories.
```

tasomaniac:
I thought I've been using it with 5.1.1 but I don't have checkstyle integrated. So something is wrong with checkstyle then.

I will have a look.

zegnus:
This is the line that seems to break:
https://github.com/novoda/gradle-static-analysis-plugin/blob/master/plugin/src/main/groovy/com/novoda/staticanalysis/internal/SourceFilter.groovy#L36

`task.source = task.source.findAll { !excludedFiles.contains(it) }`

tasomaniac:
I've seen that Gradle team had a fix for checkstyle on 5.2. Can you double check with that version? Thanks

zegnus:
Checked with `distributionUrl=https\://services.gradle.org/distributions/gradle-5.2-all.zip`, it also fails

pablisco:
Tried with 5.2.1 and Java 8 (below 8 not supported) and it assembles correctly. 
It fails on tests.
The first failure was regarding the lack of `settings.gradle` (as it was using the one from the runner project instead of the test project). That one is easy to fix.
After fixing that, we get the following error thrown on every test:
```
[...]

Issue #94, opened 2018-02-16
FindbugsConfigurator uses deprecated APIs


tasomaniac:
`sourceSet.output.classesDir` is deprecated and replaced by `sourceSet.output.classesDirs` (plural).
We should replace it to support multiple dirs.

PR #169, opened 2019-03-04
[WIP] Upgrade gradle to 5

pablisco:
This is an experiment to see if we can upgrade to Gradle 5.

Locally the tests fail so it would be interesting to see if they work on CI


zegnus:
should this go to `develop` instead?

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
```diff
@@ -27,13 +27,16 @@ ${project.additionalConfiguration}
     String toolsConfig = ''
 
     TestProject(Closure<String> template) {
+        def projectName = "${System.currentTimeMillis()}"
```
tasomaniac: These changes seem to be fine but I'm not sure if they are necessary. I would prefer to try with CI without these first and see what it says.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-all.zip
```
tasomaniac: There is a wrapper configuration in the build.gradle file. Do you mind updating that and than calling wrapper task? That way everything would be consistent. Thanks.

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
pablisco: Gradle 5 wasn't happy about not having a `settings.gradle` and we were having issues with the new memory managment (hence the gradle.properties)

tasomaniac: settings.gradle was always a requirement. Just not in tests. I'm afraid what you mention could be just for your local setup. Have you run it without settings.gradle in the CI?

Of course this can be Ok to bring in but at the same time it would be beneficial to double check with CI and see what it says.

sample/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.2.1-all.zip
```
tasomaniac: Same thing for sample 😜

pablisco:
The CI says no:
```
    * What went wrong:
    Could not create service of type FileHasher using GradleUserHomeScopeServices.createCachingFileHasher().
    > Timeout waiting to lock file hash cache (/tmp/.gradle-test-kit-jenkins/caches/5.2.1/fileHashes). It is currently in use by another Gradle instance.
      Owner PID: 22085
      Our PID: 32765
      Owner Operation: 
      Our operation: 
      Lock file: /tmp/.gradle-test-kit-jenkins/caches/5.2.1/fileHashes/fileHashes.lock
```
I've had this problem locally too. Uncertain if we are using `GradleRunner` in a way that conflicts with gradle itself (maybe because we use `withPluginClasspath()`) 🤔

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
mr-archano: I would prefer to keep the `settings.gradle` empty, as we really don't need to name the projects. I anything they should have the same name of the folder they're generated in.

pablisco:
Test are running ok now, but we have a problem which is fixed on #170 :
```
* What went wrong:
    Could not determine the dependencies of task ':findbugsDebug'.
    > Could not get unknown property 'classesDir' for debug classes of type org.gradle.api.internal.tasks.DefaultSourceSetOutput.
```

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
```diff
@@ -32,8 +32,11 @@ ${project.additionalConfiguration}
         this.gradleRunner = GradleRunner.create()
                 .withProjectDir(projectDir)
                 .withPluginClasspath()
+                .withTestKitDir(new File(projectDir, "test-kit"))
```
mr-archano: @pablisco what is this fixing?

pablisco:
From the build, we also need a `settings.gradle` in the test project generated by `DeployRulesTestRule`
```
Project directory '/storage/jenkins/workspace/gradle-static-analysis-plugin-prb/plugin/build/test-projects/1551778496887' is not part of the build defined by settings file '/storage/jenkins/workspace/gradle-static-analysis-plugin-prb/settings.gradle'. If this is an unrelated build, it must have its own settings file.
```

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
pablisco: This is for the problem of locking files while doing the internal build. Otherwise it shares the test-kit files and none of the tests work

pablisco:
@mr-archano 
The problem now is that, after changing to use to `classesDirs` FindBugs tests is broken because `project.fileTree` takes a single file instead of a list of them... So the tests that expect failures are passing (I imagine that it doesn't find any sources)

Looking at this.

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
pablisco: @mr-archano Removed the contents of the `settings.gradle` however, it was using the name already generated and used for the test project

pablisco:
Correction, it's not just `FindBugs` that is giving false positives

pablisco:
Current issue:
* What went wrong:
A problem occurred configuring root project '1551803380988'.
> Failed to notify project evaluation listener.
   > org.gradle.api.tasks.TaskInputs.dir(Ljava/lang/Object;)Lorg/gradle/api/tasks/TaskInputs;
   > org.gradle.api.tasks.TaskInputs.dir(Ljava/lang/Object;)Lorg/gradle/api/tasks/TaskInputs;

Looking into it it turns out that we have an outdated version of the Kotlin plugin which I presume didn't work with Gradle 5

tasomaniac:
🤞

pablisco:
I think that adding `.withTestKitDir(new File(projectDir, "test-kit"))` to the tests adds significant run time for the tests (from 15 mins to about an hour). So it may be worth exploring a better alternative 🤔

tasomaniac:
The CI is failing because of the deprecation in Findbugs. Aka #94. I would suggest to handle that in a separate PR using Gradle 4.10 which would make this one easier. Wdyt?

mr-archano:
@tasomaniac I suggested the same to @pablisco, but I thought there was some problem with that anyway, given the build for #170 was failing too. @pablisco could you please give us an update when you manage?

tasomaniac:
I believe the change in #170 may not be enough.

tasomaniac:
I believe the change in #170 may not be enough.

tasomaniac:
Sorry, accidentally closed

pablisco:
Got a fix on the findBugs problem: 4e5a6ed5306375956912919f0712fe38a1d8c887
```groovy
def fileTrees = sourceSet.output.classesDirs.collect { classesDir ->
    project.fileTree(classesDir) {
        builtBy(sourceSet.output)
        include(includes)
    }
}
if (fileTrees.isEmpty()) {
    throw GradleException("No file tree from ${sourceSet} and filters: ${includes}")
} else if (fileTrees.size() == 1) {
    fileTrees.first()
} else {
    fileTrees.inject(fileTrees.head()) { result, current -> result + current }
}
```
Why couldn't they be called `map` and `fold` like in other languages? 🤦‍♂️ Groovy

pablisco:
Hmm, it seems to also fix the issues with `KtlintIntegrationTest` locally, let's check with the CI in the morning 🌞

tasomaniac:
Hurray 🎉 Ktlint is failing now because of Android tooling. Updating AGP version in test projects may help.

pablisco:
Ok, it didn't work for `KtlintIntegrationTest`. But we are down to 5 failures 🎉

mr-archano:
@pablisco let's have a chat about this method, I don't know if you have checked when this was introduced, but that code was copy/pasted from the `FindbugsPlugin` source and has changed since then: https://github.com/novoda/gradle-static-analysis-plugin/pull/20#discussion_r100020596

pablisco:
The code has changed but the reason hasn't, has it?

mr-archano:
@pablisco don't know to be honest, I wonder if there's a simpler way of achieving what we want (as per description of #20)

tasomaniac:
Looked into what's wrong with ktlint. It is because of this issue https://github.com/JLLeitschuh/ktlint-gradle/issues/201 

My test upgrade caused the failure :) I guess that's for the better. Will update the tests in this branch.

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
```diff
@@ -43,6 +46,16 @@ ${project.additionalConfiguration}
         return dir
     }
 
+    void createGradleSettings() {
+        write("", 'settings.gradle')
+    }
+
+    void createGradleProperties() {
+        write("""
+org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
```
tasomaniac: isn't 2gb too much? These are almost empty projects.

pablisco: Probably, however, `xmx` referrers to maximum memory, so should be ok. Happy to lower it though.

pablisco:
Regarding the time inflation we are observing, it was suggested on Gradle's Slack a possible solution involving making a snapshot of the `testKit` and copying it over for each test...

plugin/src/test/groovy/com/novoda/test/TestProject.groovy
tasomaniac: I was just thinking this might be another reason of inflated execution time. It seems like it isn't.

pablisco:
@tasomaniac  It's still failing on 7.0 🤔 
```
An exception occurred applying plugin request [id: 'org.jlleitschuh.gradle.ktlint', version: '7.0.0']
```

tasomaniac:
I thought 7.0 bug was only for Android projects. We should then disable it completely.

mr-archano:
Closing this as partially superseded by the `0.8.1` release (#182).

The `OutOfMemoryError` issue with `test-kit` will be tracked by #168.

PR #170, opened 2019-03-04
Change deprecated classesDir for classesDirs

### LINKED ISSUE #94: FindbugsConfigurator uses deprecated APIs

pablisco:
`classesDir` is deprecated in Gradle 4.x and we should use `classesDirs` instead as Gradle 5.x removes this property.

Without this change, Gradle complains about it if we use Gradle 5.x on the client:
```
* What went wrong:
Could not determine the dependencies of task ':core-java:findbugsMain'.
> Could not get unknown property 'classesDir' for main classes of type org.gradle.api.internal.tasks.DefaultSourceSetOutput.

```

Fixes #94


tasomaniac:
This is duplicate of #97 but I am closing #97 since you nabrucj more details here.

tasomaniac:
The fix for this is easy actually. We should just use classesDirs instead.

tasomaniac:
Silly me. I thought this was an issue. Didn't realize it was a PR 🤦‍♂️ haha 😂

I couldn't see why the CI was failing. There might be multiple usages of `classesDir`

pablisco:
Moved into #169 as part of the upgrade

PR #171, opened 2019-03-06
[TEST] Test no inline test kit

pablisco:
This extends from #169 and it's a mere test to see if we need to inline test kit on the CI


tasomaniac:
Let's maybe try this one more time because this might be failing because they were run together

pablisco:
Yeah, that was my thought so, on the second commit, I moved `testKit` to a build local folder

PR #172, opened 2019-03-07
Change deprecated classesDir for classesDirs

pablisco:
Note: copy of #170 due to rebase

`classesDir` is deprecated in Gradle 4.x and we should use `classesDirs` instead as Gradle 5.x removes this property.

Without this change, Gradle complains about it if we use Gradle 5.x on the client:
```
* What went wrong:
Could not determine the dependencies of task ':core-java:findbugsMain'.
> Could not get unknown property 'classesDir' for main classes of type org.gradle.api.internal.tasks.DefaultSourceSetOutput.

```

Fixes #94

PR #174, opened 2019-03-08
Change deprecated classesDir for classesDirs

### LINKED ISSUE #94: FindbugsConfigurator uses deprecated APIs

tasomaniac:
Copy of #172 

Fixes #94

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
@@ -126,7 +126,7 @@ class FindbugsConfigurator extends CodeQualityConfigurator<FindBugs, FindBugsExt
      * being passed to FindBugs Ant task, resulting in an error
      * */
     private ConfigurableFileTree createClassesTreeFrom(SourceSet sourceSet) {
-        project.fileTree(sourceSet.output.classesDir, {
+        project.fileTree(sourceSet.output.classesDirs, {
             it.builtBy(sourceSet.output)
         })
     }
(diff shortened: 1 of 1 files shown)


plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
```diff
@@ -118,17 +120,27 @@ class FindbugsConfigurator extends CodeQualityConfigurator<FindBugs, FindBugsExt
...
-        })
+    private FileTree createClassesTreeFrom(SourceSet sourceSet, List<String> includes) {
+        def fileTrees = sourceSet.output.classesDirs.collect { classesDir ->
+            project.fileTree(classesDir) {
+                builtBy(sourceSet.output)
+                include(includes)
+            }
+        }
+        if (fileTrees.isEmpty()) {
+            throw GradleException("No file tree from ${sourceSet} and filters: ${includes}")
```
mr-archano: I was thinking about this and I am not sure we want to be this strict here...

plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
```diff
@@ -118,17 +120,27 @@ class FindbugsConfigurator extends CodeQualityConfigurator<FindBugs, FindBugsExt
...
+        } else if (fileTrees.size() == 1) {
+            fileTrees.first()
+        } else {
+            fileTrees.inject(fileTrees.head()) { result, current -> result + current }
```
mr-archano: I believe we should be able to `inject` while iterating on `classesDirs`, this code could be a lot simpler in the end, something like [this](https://github.com/novoda/gradle-static-analysis-plugin/blob/master/plugin/src/main/groovy/com/novoda/staticanalysis/internal/SourceFilter.groovy#L40):
```
classesDirs.inject(null) { tree, file -> tree?.plus(project.fileTree(file)) ?: project.fileTree(file) }.include(includes)
```

tasomaniac:
Merge merge merge!!!

PR #177, opened 2019-03-09
Make test-projects runnable without test-kit

mr-archano:
### Problem

When a test-project fails is not possible at the moment just running that single project again, because of the way `GradleRunner` is using the `java-gradle-plugin` to inject the plugin into the classpath of the build.

### Solution

Simply get rid of `java-gradle-plugin` and use composite builds instead. In order to achieve that I have introduced a Gradle project including the `main` sourceset of `plugin`. This project is then used by the test projects using a simple `dependencySubsitution` rule in their `settings.gradle`.


mr-archano:
Yes it does, I already checked :)

PR #183, opened 2019-04-13
Remove afterEvaluate together with support for older versions of Ktlint plugin

### LINKED ISSUE #156: Remove project.afterEvaluate in KtlintConfigurator
`project.afterEvaluate` was needed for the older versions of Ktlint. 

Now it is not needed. It is not so easy to conditionally put it whenever an older version is used. So, I'm suggesting to just remove it. If removed versions 6.+ would be supported by default.

Users still using the old version would need to put `afterEvaluate` themselves. I guess we can make it clear on release notes.

tasomaniac:
`afterEvaluate` usage is not good since it requires the clients to use `afterEvaluate` in certain conditions. 

This was required initially because ktlint plugin itself was using it. That is fixed quite a while ago. 

## Solution: 

- Remove `afterEvaluate`
- Remove older version support for ktlint
- Update the docs to indicate that we support version 6 and above

Fixes #156

### DIFF
--- a/docs/tools/ktlint.md
+++ b/docs/tools/ktlint.md
@@ -2,6 +2,8 @@
 [Ktlint](https://github.com/shyiko/ktlint) is a linter for Kotlin with a built-in formatter. It does not support Java. Adding 
 this tool only makes sense when you have Kotlin sources in your project. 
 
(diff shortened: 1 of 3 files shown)


tasomaniac:
Yes, just fixed it. 🙌

PR #184, opened 2019-05-04
Introduce lazy tasks except for ktlint

tasomaniac:
Finally finished with lazy tasks except for ktlint. 

Ktlint is more challenging since we support multiple versions. That's why we need to resolve tasks to identify. 

For Checkstyle, PMD and Findbugs, I've changed `CodeQualityConfigurator` a lot. How we configure tasks is completely redesigned. 
- Now we create `collectViolations` tasks per sourceSet
- Then `collectViolationsVariant` as a meta task to trigger others. This one is also used for variantFilter.

Android Lint was easier. Just used `register` instead of `maybeCreate`. But I had to hold on a flag to make sure that the task is created only once. There is no equivalent `maybeRegister` in Gradle.

 Detekt is also similar. But now, I've realized that we don't support configuring multiple times. I will do a follow-up to support that.

| without anything | with Checkstyle, Pmd, Findbugs, Lint |
| -- | --- |
| image | image |

Notice that the number of tasks configured is the same (only 23)

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
@@ -62,7 +62,7 @@ abstract class CodeQualityConfigurator<T extends SourceTask & VerificationTask, 
         }
     }
 
-    def configureAndroidWithVariants(DomainObjectSet variants) {
+    protected void configureAndroidWithVariants(DomainObjectSet variants) {
         project.android.sourceSets.all { sourceSet ->
             createToolTaskForAndroid(sourceSet)
             createCollectViolations(getToolTaskNameFor(sourceSet), violations)
@@ -72,7 +72,7 @@ abstract class CodeQualityConfigurator<T extends SourceTask & VerificationTask, 
         variantFilter.filteredUnitTestVariants.all { configureVariant(it) }
     }
 
-    protected configureVariant(variant) {
+    protected void configureVariant(variant) {
         def collectViolations = createVariantMetaTask(variant)
         evaluateViolations.dependsOn collectViolations
     }
@@ -104,9 +104,9 @@ abstract class CodeQualityConfigurator<T extends SourceTask & VerificationTask, 
         }
     }
 
-    private static def javaCompile(variant) {
+    protected static def javaCompile(variant) {
         if (variant.hasProperty('javaCompileProvider')) {
-            variant.javaCompileProvider
+            variant.javaCompileProvider.get()
         } else {
             variant.javaCompile
         }
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
@@ -99,9 +99,11 @@ abstract class CodeQualityConfigurator<T extends SourceTask & VerificationTask, 
 
     protected abstract String getToolName()
 
-    protected Object getToolPlugin() {
+    protected def getToolPlugin() {
         toolName
     }
+
+    protected abstract Class<T> getTaskClass()
 
     protected abstract Class<E> getExtensionClass()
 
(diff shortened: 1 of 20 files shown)


tasomaniac:
Guess running too many PRs ws too much for the machine. Gradle daemon disappeared. Can someone retry the task please?

plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
```diff
@@ -78,17 +111,29 @@ abstract class CodeQualityConfigurator<T extends SourceTask, E extends CodeQuali
...
+        "$toolName${sourceSet.name.capitalize()}"
+    }
+
     protected abstract Class<T> getTaskClass()
 
-    protected abstract void configureReportEvaluation(T task, Violations violations)
+    protected abstract void createToolTaskForAndroid(sourceSet)
+
+    protected void configureToolTask(T task) {
+        sourceFilter.applyTo(task)
```
tasomaniac: Collected all common functionality here. It is automatically applied to all tool tasks thanks to `configureEach` in line 53 above.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
```diff
@@ -78,17 +111,29 @@ abstract class CodeQualityConfigurator<T extends SourceTask, E extends CodeQuali
...
-            task.exclude '**/*.kt'
+    protected static def javaCompile(variant) {
+        if (variant.hasProperty('javaCompileProvider')) {
+            variant.javaCompileProvider.get()
+        } else {
+            variant.javaCompile
         }
     }
 
+    protected final String getToolTaskNameFor(sourceSet) {
```
mr-archano: _technically_ this isn't always a sourceset right? maybe we can just called it `named`?

plugin/src/main/groovy/com/novoda/staticanalysis/internal/TasksCompat.groovy
```diff
@@ -0,0 +1,37 @@
...
+
+import org.gradle.api.Action
+import org.gradle.api.Project
+import org.gradle.api.Task
+import org.gradle.api.tasks.TaskCollection
+import org.gradle.util.GradleVersion
+
+class TasksCompat {
+
+    private static boolean IS_GRADLE_MIN_49 = GradleVersion.current() >= GradleVersion.version("4.9")
```
mr-archano: `nit:` I'd prefer `'` over `"` 💃

plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
```diff
@@ -78,17 +111,29 @@ abstract class CodeQualityConfigurator<T extends SourceTask, E extends CodeQuali
...
+
     protected abstract Class<T> getTaskClass()
 
-    protected abstract void configureReportEvaluation(T task, Violations violations)
+    protected abstract void createToolTaskForAndroid(sourceSet)
+
+    protected void configureToolTask(T task) {
+        sourceFilter.applyTo(task)
+        task.group = 'verification'
+        task.exclude '**/*.kt'
```
mr-archano: is this ok because this configurator is bound to `SourceTask` & `CodeQualityExtension`?

plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
tasomaniac: Soooo, this was actually only for `sourceSet`. Then I realized that `FindBugs` needs to be configured per variant because it needs `javaCompile` task. I will go over and make sure that naming is consistent.

plugin/src/main/groovy/com/novoda/staticanalysis/internal/CodeQualityConfigurator.groovy
tasomaniac: Yes, here `exclude` method comes from `SourceTask`.

PR #185, opened 2019-05-08
Bug fix: define closure strategy

tasomaniac:
# Problem

We are using very powerful feature of Closure in Groovy which is to define delegates. I've realized that some properties were just not working if they clash with something that is already defined in my script.

If it is a function call, it would call the function I defined instead of the one in the tool.

# Solution

I've researched and learned that Closure's prefer the Owner over the Delegate first. But by providing custom strategy, we can force to use only the delegate. 

In this case, the delegate is the static analysis tool we are integrating.


tasomaniac:
The one I came across is `version`. It's a property in ktlint but when used, it sets the `version` property of the Gradle project instead.

tasomaniac:
Here is what I've just tried in the `sample` project

```

void toolVersion(String foo) {
    throw new IllegalStateException()
}

staticAnalysis {

    penalty none

    checkstyle {
        toolVersion '8.8'
    }
}
```

Configuring the project fails with `IllegalStateException` since `toolVersion` inside checkstyle calls the wrong method defined above.

image

PR #188, opened 2019-05-15
Update samples

tasomaniac:
- Update sample dependencies to latest versions to be able to showcase lazy task configuration
- Renamed `buildpropertiesplugin` copy pasta :) 

- Fix a small issue with Findbugs
  - I think this was not picked up by tests because it is Gradle 5 behavior change. 
  - I got `Task with name 'collectFindbugsDebugUnitTestViolations' not found` while configuring html generation task for findbugs.


sample-multi-module/app/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,6 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
   xmlns:tools="http://schemas.android.com/tools"
-  package="com.novoda.buildpropertiesplugin.sample">
```
mr-archano: 🦊

PR #189, opened 2019-05-16
Lazy Task Configuration for ktlint

### LINKED ISSUE #161: Support task configuration avoidance
Did some benchmarking for task configuration avoidance. 
This was done on [Open Link With](https://github.com/tasomaniac/OpenLinkWith) using Android Gradle Plugin 3.4.0-alpha08 and Gradle 5.1-rc-1.

| Without SAP | With SAP |
| ------------ | --------- |
| image | image |

As can be seen from above **368 tasks** in total had to be configured when SAP is applied. It must almost all of the Android tasks. 
 
Here is a quite nice example of how to support task avoidance by keeping backward compatibility. https://github.com/melix/jmh-gradle-plugin/blob/a034aa88805b7a06fa9c5a825d573554b2aa23e2/src/main/groovy/me/champeau/gradle/JMHPlugin.groovy#L289-L296

tasomaniac:
This is the last PR of the series where I enable lazy task configuration. Fixes #161 

We had somewhat complex (but well tested) setup for ktlint where we support multiple versions even-though there were breaking changes in between. To be able to do this, I used `matching` to dynamically load all Ktlint tasks and pick the ones that we are interested in. This of course resolve tasks and breaks task configuration avoidance. 

In order to handle this, we need to know more about how Ktlint tasks are created. With the latest versions, they are created per source set which is pretty consistent with other tools configurations we have 

This is why support for version 6.1 is removed.

### Additionally 

Also added try-catch for any future exceptions with a clear output of the current state with clear message of which versions of Ktlint and Detekt we support.

### Final screenshot: 

All tools enabled from `sample` and majority of the tasks are avoided 

image

### DIFF
--- a/docs/tools/ktlint.md
+++ b/docs/tools/ktlint.md
@@ -2,7 +2,7 @@
 [Ktlint](https://github.com/shyiko/ktlint) is a linter for Kotlin with a built-in formatter. It does not support Java. Adding 
 this tool only makes sense when you have Kotlin sources in your project. 
 
-> Supported Ktlint Gradle Plugin version: **6.0.0 and above** 
+> Supported Ktlint Gradle Plugin version: **6.2.1 and above** 
 
 ## Table of contents
  * [IMPORTANT: setup Ktlint](#important-setup-ktlint)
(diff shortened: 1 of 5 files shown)


tasomaniac:
Failed because Daemon disappeared unexpectedly. This happens when CI is busy/loaded. Will retry later

mr-archano:
@hal9002 retest this please

mr-archano:
@hal9002 retest this please

tasomaniac:
All green 👍

PR #194, opened 2019-05-30
Docs/Update custom evaluator sample

lgvalle:
This PR:

* Adds a more elaborated code sample for custom evaluators
* Fix the previous sample code (as it won't compile as it was)
* Fix the link path to `Violations` source code

### DIFF
--- a/docs/incubating/custom-evaluator.md
+++ b/docs/incubating/custom-evaluator.md
@@ -40,7 +40,7 @@ can be provided as a closure as well:
 
 ```gradle
 staticAnalysis {
-    evaluator { Set<Violations> allViolations ->
+    evaluator { Set allViolations ->
        // add your evaluation logic here
     }
     //...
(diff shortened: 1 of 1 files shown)


docs/incubating/custom-evaluator.md
```diff
@@ -61,10 +61,29 @@ Anything that respect such contract is valid. For example, a custom evaluator mi
  * Only break the build if there are errors or warnings in one specific report
  * Or anything else that you can think of
 
+For example, this custom evaluator fails the build if PMD errors are greater than five:
+
+```gradle
+evaluator { Set allViolations ->
+    allViolations.each { violation ->
+        if (violation.name == "PMD" && (violation.errors > 5)) {
```
tasomaniac: Below it says `toolName` but here it is name. Which one it is? 🤔

lgvalle: Good point! The backing property is called `toolName` but the method to access it is `getName` (and I'm _guessing_ that's why the compiler allows `.name` as accessor?) 

Do you think we shall stick to just `name` in the sample to create less confusion?

tasomaniac: Looks like `toolName` is private. We shouldn't really mention that since it is not usable. Sticking to `name` makes sense 👍

lgvalle: amended here d3da57b

PR #197, opened 2019-08-27
Update Gradle to 5.6

zegnus:
- Upgrade Gradle to [5.6](https://docs.gradle.org/5.6/release-notes.html)

This solves https://github.com/novoda/gradle-static-analysis-plugin/issues/168


mr-archano:
@zegnus the CI is not happy about the change. Could you please look into it and let us know what is the problem?

zegnus:
@mr-archano the CI runs out of memory... I honestly do not know how to fix this :(

mr-archano:
@zegnus have you tried with this? https://docs.gradle.org/current/userguide/build_environment.html#sec:configuring_jvm_memory

Gradle 5.x has changed the amount of memory allocated by default to the daemon. It could be also worth be explicit in the amount of memory assigned to the test process (see https://docs.gradle.org/current/dsl/org.gradle.api.tasks.testing.Test.html)

tasomaniac:
Superseeded by #203

Issue #201, opened 2019-10-13
Unable to support JLLeitschuh/ktlint-gradle plugin version 9.0.0


AdamMc331:
I recently updated to version 9.0.0 of the [ktlint-gradle](https://github.com/JLLeitschuh/ktlint-gradle) plugin and my builds started failing with this stack trace:

```
Caused by: groovy.lang.MissingPropertyException: Could not get unknown property 'reportOutputFiles' for task ':app:ktlintTestDebugSourceSetCheck' of type org.jlleitschuh.gradle.ktlint.KtlintCheckTask.
        at org.gradle.internal.metaobject.AbstractDynamicObject.getMissingProperty(AbstractDynamicObject.java:84)
        at org.gradle.internal.metaobject.AbstractDynamicObject.getProperty(AbstractDynamicObject.java:61)
        at org.jlleitschuh.gradle.ktlint.KtlintCheckTask_Decorated.getProperty(Unknown Source)
        at com.novoda.staticanalysis.internal.ktlint.KtlintConfigurator$_createCollectViolationsTask_closure9.doCall(KtlintConfigurator.groovy:137)
```

Then I noticed, in the latest changelog, this task was replaced by `allReportsOutputFiles`: https://github.com/JLLeitschuh/ktlint-gradle/blob/master/CHANGELOG.md#added-1

But this plugin appears to still look for `reportOutputFiles`: https://github.com/novoda/gradle-static-analysis-plugin/blob/353752cd85c116f6c7f015f61c9c638b20a5e897/plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy#L137

I don't think it's as simple as just updating this plugin because it may break for people who don't want to _also_ update the KtLint gradle plugin. 

What are some options we have? Let me know if I can help contribute a solution.

tasomaniac:
That's one of the reasons why we don't update this plugin to Kotlin. Groovy is more dynamic.

We support multiple versions of plugins by checking availability of certain properties. The solution is to check if the property is available. One example is available here:

https://github.com/novoda/gradle-static-analysis-plugin/blob/353752cd85c116f6c7f015f61c9c638b20a5e897/plugin/src/main/groovy/com/novoda/staticanalysis/internal/detekt/DetektConfigurator.groovy#L81

AdamMc331:
[...]

PR #202, opened 2019-10-20
Updated KtLintConfigurator to check for new report file name for #201.

AdamMc331:
Currently this plugin won't support version 9.0.0+ from the ktlint-gradle plugin: #201 

This PR resolves it by:

- Adds version 9.0.0 to the integration tests, and updated them to make sure we were showing the right config for the version.
- Added an if/else statement that handles the renamed property, and will now also throw a more relevant error if it were to change again.
- Updated the gradle version to 5.x because that's required by the latest ktlint-gradle plugin. 
- Updated the sample app to point to version 9.0.0

Version 5.6.2 might be a larger jump than necessary, but I wasn't sure how to determine what a good change was. For ktlint-gradle the minimum supported version is `5.4.1`, so let me know if we should use that instead. 

Validated by updating to 9.0.0 in the sample app and ensuring that runs, as well as running all of the tests locally.


hal9002:
Can one of the admins review this PR?

plugin/src/main/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintConfigurator.groovy
```diff
@@ -134,7 +136,21 @@ Last tested compatible version: $LAST_COMPATIBLE_KTLINT_VERSION
...
             File txtReportFile = null
             try {
-                ktlintTask.reportOutputFiles.each { key, fileProp ->
+                Map<String, RegularFileProperty> outputFiles = null
+
+                /**
+                 * This change was made to support version 9.0.0 of the JLLeitschuh/ktlint-gradle plugin which renamed
+                 * this property: https://github.com/JLLeitschuh/ktlint-gradle/blob/master/CHANGELOG.md#900---2019-09-30
+                 */
+                if (ktlintTask.hasProperty("allReportsOutputFiles")) {
```
tasomaniac: Because of its nature, this code class is not the cleanest. We can help it more clean by moving this into a private function below. The comment can go above the function then. What do you think?

plugin/src/test/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintIntegrationTest.groovy
```diff
@@ -124,4 +137,12 @@ class KtlintIntegrationTest {
                     maxErrors = ${maxErrors}
                 }""")
     }
+
+    private String getDefaultConfig() {
```
tasomaniac: Just an optional suggestion: 

Since this test is already parameterized, this `defaultConfig` can also be another property on the parameters. Thinking about this also because of `>=` on the `String` type below.

sample/app/build.gradle
```diff
@@ -66,7 +66,9 @@ staticAnalysis {
 
     ktlint {
         android.set(true)
-        reporters = [ReporterType.CHECKSTYLE]
```
tasomaniac: In order to do this, I had to put `ReporterType` in some other Gradle file. That was hacky. Glad that they fixed this. Can you search for `RegularFileProperty` in the sample folder and remove it? Thanks.

tasomaniac:
@hal9002 ok to test

tasomaniac:
ok to test

plugin/src/test/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintIntegrationTest.groovy
AdamMc331: I like that a lot better. Will make any future changes to the config feel much cleaner too.

tasomaniac:
Here goes my failed attempts to trigger CI checks. @mr-archano can you help?

AdamMc331:
@tasomaniac I've made your suggested changes. I've also noticed that my git history looks a little weird, and I think that's because on my fork I rebased off of master and not develop.

Let me know if you want me to correct this & rebase off of develop, and I can do that once we've reviewed these changes.

tasomaniac:
Also @AdamMc331, can you rebase into develop and only keep your commit(s)? The default branch is `master` because we want people to see the docs in master.

AdamMc331:
All set! I didn't squash mine but let me know if that's preferred and I can do this again. 🙈

Mecharyry:
Test this please

plugin/src/test/groovy/com/novoda/staticanalysis/internal/ktlint/KtlintIntegrationTest.groovy
tasomaniac: Awesome. Thanks

tasomaniac:
CI is running. Fingers crossed 🤞 As we tried #197 before, we had memory issues. Now i'm also trying here #203. With #203, I did some optimizations on tests to consume less memory. If that goes well, we can merge both and do a release. 

Btw, with this change, our Gradle requirement does not change. Since we don't have hard-dependency to ktlint, we will still be supporting Gradle 4.10 and above.

AdamMc331:
@tasomaniac Are these the same memory related issues you've seen before?

tasomaniac:
Yep. Finally had success in #203. Now I need a review/approval. Will ping @mr-archano for that and we can release this soon.

AdamMc331:
Great, once that is resolved just ping me and I will rebase this again.

tasomaniac:
Thanks to @Mecharyry it is actually done. Merging once CI is good.

tasomaniac:
Test this please

Mecharyry:
Test this please oh mighty CI, destroyer of dreams

AdamMc331:
@tasomaniac Looks like we're all good! Thanks so much for your review and help on this. :)

tasomaniac:
Thank you for the contribution. 🙌 Prepared PR for releasing version 1.1 #206

PR #203, opened 2019-10-21
Another try with gradle update

tasomaniac:
Compared to what we did before, here is what I'm doing differently:

- ~Trying with Gradle 6 to see if they have anything differently.~ Looks like Findbugs is completely removed from Gradle 6. Trying with 5.6.2 now
- I realized that we run lint with every other check. Introduced a conditional to disable lint for non-lint related tests.


tasomaniac:
Wow!! 🎉 It passed even quicker than the last task on Gradle 4.10

PR #213, opened 2019-11-08
Fix kapt issue

### LINKED ISSUE #199: Gradle error with Kapt
We've observed a problem when running `gradle-static-analysis-plugin` alongside `kapt`:

```
* What went wrong:

Could not determine the dependencies of task ':models:findbugsMain'.

> No signature of method: com.novoda.staticanalysis.internal.findbugs.FindbugsConfigurator$_createClassesTreeFrom_closure11.doCall() is applicable for argument types: (org.gradle.api.internal.file.UnionFileTree, File) values: [file tree, /storage/jenkins/workspace/ALL-4-PIPELINE_PR-10233/base/models/build/tmp/kapt3/classes/main]

  Possible solutions: doCall(org.gradle.api.file.ConfigurableFileTree, java.io.File), findAll(), findAll()
```

We have `@AutoValue` set up using `kts` like this:

```
plugins {
  // ...
  kotlin("kapt")
}

dependencies {

  // ...

  compileOnly("com.google.auto.value:auto-value-annotations:1.6.3") 
  kapt("com.google.auto.value:auto-value:1.6.3")

}

```

The problem is solved by removing `kotlin("kapt")` and changing the `kapt` dependency to `annotationProcessor`

tasomaniac:
Use Simple FileTree/FileCollection instead of ConfigurableFileTree

We were never dependant on that particular class and it seems like applying other plugins change the implementation class.

## Tests

Also added tests to first reproduce it and then fix it.

For this, created new suite of tests for java-only plugins to also include Kotlin projects and test configuration. 

Fixes #199

### DIFF
--- a/plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
+++ b/plugin/src/main/groovy/com/novoda/staticanalysis/internal/findbugs/FindbugsConfigurator.groovy
@@ -2,14 +2,13 @@ package com.novoda.staticanalysis.internal.findbugs
 
 import com.novoda.staticanalysis.Violations
 import com.novoda.staticanalysis.internal.CodeQualityConfigurator
-import com.novoda.staticanalysis.internal.CollectViolationsTask
 import org.gradle.api.Action
 import org.gradle.api.DomainObjectSet
 import org.gradle.api.NamedDomainObjectContainer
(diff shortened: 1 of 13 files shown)


tasomaniac:
I'm going to merge this. I tagged @pablisco before but it is not strictly necessary for him to review. He can instead verify the fix later. I reproduced it with tests and verified the fix anyways.

PR #218, opened 2019-11-21
Fix Gradle 6 issue

### LINKED ISSUE #204: Prepare for Findbugs removal in Gradle 6.0
Gradle removed Findbugs in 6.0. People may not realize that once they upgraded. First thing that might fail is actually our plugin if they don't use findbugs directly. 

So a proper error message would be useful. Worst-case we should add this into our readme.

They will need to remove findbugs in order to solve this issue. Nothing else can really be done if the user wants to upgrade to 6.0

tasomaniac:
Gradle loads classes with all its dependencies. So FindbugsConfigurator.create would fail on Gradle 6. Guarding it in a separate Factory class solves the issue

Fixes #204


tasomaniac:
Opened towards master to be able to make it easier. Consider this as hotfix. 

Let's do the 1.2 release after this. And then we will need to merge master -> develop anyways.

mr-archano:
@zegnus I am going to merge this, but then we will need to trigger the CI job to actually release 1.2.

